PR #2, opened 2018-10-16
Fix Typo

websupporter:
`$title = esc_html__( 'Dayly Archives:', 'twentynineteen' );`
Should be 'Daily'


kjellr:
Looks good. Thanks for the fix, @websupporter!

PR #3, opened 2018-10-16
Better SEO for site branding

audrasjb:
Let's add a condition in site-branding.php since we doesn't want to display H1 tag if we are not on the homepage (for both blog page or front page).

Cheers,
Jb


2ndkauboy:
More importantly, we should do it for a better a11y ;)

kopepasah:
This looks good to me. 👍

audrasjb:
Thanks @joyously 
I updated my patch and I also added some SASS fixes to keep the `<h1>` styling when we are in a `<p>` context. Looks good to me now.

PR #9, opened 2018-10-16
Change license uri

B-07:
As the license file is included in the theme. This was also done in _s


B-07:
I'll refresh it and create a new one

PR #10, opened 2018-10-16
Create CONTRIBUTING.md

B-07:
Contributing Guidelines for Twenty Nineteen Theme


kjellr:
Looks good, thanks @B-07. 

For reference, this is basically the same as [the one in Twenty Seventeen](https://github.com/WordPress/twentyseventeen/blob/master/CONTRIBUTING.md) (which still seems appropriate).

PR #16, opened 2018-10-16
Fix indentation

B-07:
As same as other files

### DIFF
--- a/archive.php
+++ b/archive.php
@@ -51,4 +51,4 @@ get_header(); ?>
 	</div><!-- .content-area -->
 
 <?php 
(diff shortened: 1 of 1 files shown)


2ndkauboy:
You should probably remove the closig PHP tag (as in the other files).

B-07:
@2ndkauboy Yes! Fixed! Thanks!

kjellr:
@B-07 — I think this was taken care of via #167. Can you confirm?

B-07:
Yes, Looks fine! Closing this.

PR #18, opened 2018-10-16
Add ending comments

themeroots:
As all others have the same


kjellr:
Thanks, @themeroots. I'm not actually sure this document needs many of these inline comments... at least, it doesn't seem to need them all to render on the frontend. It appears that some of these come from `_s`, but a few more have been added at some point. @allancole do you know why these are here/have a preference for them?

crunnells:
These are not necessary. There is definitely a case for adding HTML comments after long blocks of code, or when closing one or more elements within a block of code. However, in both of the cases in this PR, the code is less than 5 lines so there _should_ be no confusion about which element is being closed. 

I'd recommend closing this PR, as it's not really necessary.

PR #19, opened 2018-10-16
show .site-description only if $description is not empty

### LINKED ISSUE #17: show .site-description only if $description is not empty
At the moment .site-description in template-parts/header/site-branding.php is always visible. On is_home() or is_front_page() it displays the description of the website, however on other pages it will only display the seperator (&mdash;)

enodekciw:
fixes #17 

wp.org username: enodekciw

### DIFF
--- a/template-parts/header/site-branding.php
+++ b/template-parts/header/site-branding.php
@@ -14,9 +14,11 @@
 			if ( ! preg_match( '/\.\s*$/', $description )  ) $description .= '.';
 		}
 	?>
-	<p class="site-description">
-		<span class="separator">&mdash;</span> <?php echo $description; ?>
-	</p>
+	<?php if( ! empty( $description ) ) : ?>
+		<p class="site-description">
+			<span class="separator">&mdash;</span> <?php echo $description; ?>
+		</p>
+	<?php endif; ?>
 	<?php if ( has_nav_menu( 'menu-1' ) ) : ?>
 		<nav id="site-navigation" class="main-navigation" role="navigation" aria-label="<?php esc_attr_e( 'Top Menu', 'twentynineteen' ); ?>">
 			<?php wp_nav_menu( array(


spencerfinnell:
Would this make more sense as a template tag now? There's a decent amount of logic to determine the visibility which could be further filtered in a function.

template-parts/header/site-branding.php
```diff
@@ -14,9 +14,11 @@
 			if ( ! preg_match( '/\.\s*$/', $description )  ) $description .= '.';
 		}
 	?>
-	<p class="site-description">
-		<span class="separator">&mdash;</span> <?php echo $description; ?>
-	</p>
+	<?php if( ! empty( $description ) || ( is_customize_preview() && ! is_null( $description ) ) ) : ?>
```
westonruter: The `! is_null( $description )` condition should be removed. When in the Customizer, the tagline can be supplied from nothing. Assuming `postMessage` transport is used to preview the change to the description, it needs an element in the DOM to update. 

Aside: The description should be wrapped in an element to facilitate this as well.

enodekciw: if we remove `! is_null( $description )` description becomes visible on other pages (except the ones that matches `is_home()` and `is_front_page()`) while using the customizer (see it in action: open the customizer then navigate to a post or page). Do we really want it to be editable on, let's say, `single.php` even if it's not gonna be visible there after changes are published?

westonruter: Refer to what is done in Twenty Seventeen: https://github.com/WordPress/wordpress-develop/blob/5d92e0c0a950399c53294a41000305ddcf47efb5/src/wp-content/themes/twentyseventeen/template-parts/header/site-branding.php#L24-L30

The problem here for Twenty Nineteen is the presence of `<span class="separator">&mdash;</span>`. This could be solved if the dash were added in CSS instead like as follows:

```css
.site-description:not(:empty):before {
    content: "— "
}
```

enodekciw: Yup, I've noticed that mdash and dot (which is added automatically) disappear after editing site description in customizer. But, in my opinion, that's a different issue and has nothing to do with `! is_null( $description )` check since it is needed anyway to see if we're not customizing pages other than `is_home()` or `is_front_page()` ([see this](https://github.com/WordPress/twentynineteen/blob/7b1170c700b51ccc89479a5552c97102832b2d77/template-parts/header/site-branding.php#L10))

I could work on a disappearing mdash / dot thing but I'd prefer doing that in a different PR.

enodekciw:
new PR: #129

Issue #13, opened 2018-10-16
Post Navigation aligned wrong


fabiankaegy:
The `.nav-links` element on the Single Blog has a left alignment similar to `.alignwide` blocks. But the right right side has the standard alignment. 

<img width="1808" alt="bildschirmfoto 2018-10-16 um 16 27 28" src="https://user-images.githubusercontent.com/20684594/47023706-81d3b100-d160-11e8-9eeb-bf50afbd51ec.png">

kjellr:
Closed via #21. Thanks!

PR #21, opened 2018-10-16
fixing post links alignement on single blog page

fabiankaegy:
As mentioned in issue #13 
bildschirmfoto 2018-10-16 um 16 43 47


kjellr:
Thanks for the fix, @fabiankaegy!

PR #22, opened 2018-10-16
Update editor alignment to match front end

kjellr:
The front end of the theme left-aligns all content, and lets wide width blocks extend beyond the text column on the right. We need to match the front end to the back end as much as possible, so this PR is a rough first stab at getting that alignment in place. 

- Gutenberg currently auto-prefixes editor block stylesheets. That won't work for us here since these styles need to effect the frame of the page, not just individual blocks. Until a solution for that it sorted out, this introduces a second editor stylesheet. (This also allows us to customize the page title.)
- So far, this doesn't effect any items that are aligned to the right. Those should end up hanging off the right of the text column too. 
- Once we have this working on the back end, we'll likely want to post the markup over to the frontend so it matches. 

_Before:_ 
screen shot 2018-10-16 at 10 41 00 am

_After:_
screen shot 2018-10-16 at 10 40 30 am

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -130,6 +130,15 @@ function twentynineteen_scripts() {
 add_action( 'wp_enqueue_scripts', 'twentynineteen_scripts' );
 
 /**
+ * Enqueue supplemental block editor styles
+ */
+function twentynineteen_editor_frame_styles() {
+    wp_enqueue_style( 'twentynineteen-editor-frame-styles', get_theme_file_uri( '/style-editor-frame.css' ), false, '1.0', 'all' );
+}
+
+add_action( 'enqueue_block_editor_assets', 'twentynineteen_editor_frame_styles' );
+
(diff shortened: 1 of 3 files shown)


kjellr:
Got this working with fairly simple markup in the end. 🙌 

Here's a video of it in action:
https://cloudup.com/cXuFNYeHJ9o

This only kicks in at breakpoints larger than `768px`. Otherwise, everything's just a single column anyway. The main caveat for this whole PR is just the addition of a separate editor stylesheet, but that's necessary until we can customize the page title, editor frame margins, and appender text within `style-editor.css` (we can't right now because of Gutenberg's auto-prefixing).

I gave this a spin in every browser I've got (Safari, Chrome, FF, IE11, and mobile Safari) but this could use some additional testing just to be sure it's working alright.

PR #38, opened 2018-10-16
Improve search container on 404 template

### LINKED ISSUE #37: Fix: Search container on 404 template can have margins
## Current search container on a 404 page

On a 404 page, 

- The search button seems to be squished to the search field
- The overall search container is missing margins as well, given it a look as shown below

screenshot 2018-10-16 at 22 26 16

## Suggested view

Something like this looks better.

screenshot 2018-10-16 at 22 27 12

arunsathiya:
## Description

Resolves #37

- Introduces top margin for the search button so that the button is no longer squished with the search field
- Introduces margin to the overall search container

#### Before

screenshot 2018-10-16 at 22 26 16

#### After

screenshot 2018-10-16 at 22 27 12

---

🤞🏽 First ever WordPress PR, fingers crossed this looks right!

**WordPress.org username:** arunsathiya


nielslange:
@arunsathiya While adding the code to the *.css file fixes this issue on (very) short term, you should add the fix to one of the *.sass files, to persist it. 😉

PR #42, opened 2018-10-16
Update theme and license URI

B-07:
License URI changed because license file is included in the theme, same was done in _s


kjellr:
Thanks, @B-07!

PR #54, opened 2018-10-16
🐛 FIX: Missing calc Bracket

asharirfan:
Added a missing `calc` bracket that caused an error during a SASS compile test.

WP.org username: mrasharirfan


kjellr:
Looks good, thanks @asharirfan!

PR #55, opened 2018-10-16
❌ REMOVE: Jetpack Infinite Scroll Styles

### LINKED ISSUE #44: Jetpack Infinite Scroll Styles Not Needed
I see some styles for Jetpack infinite scroll functionality. That's usually not done in default themes, but instead handled in Jetpack itself. So those can be removed.

asharirfan:
Fixes #44 

This PR is to remove Jetpack infinite scroll styles from the theme.

WP.org username: mrasharirfan


allancole:
Good call. Merging this in now ;-)

westonruter:
What about:

https://github.com/WordPress/twentynineteen/blob/ebf80e929d3f87014383cb279a14132282373f2a/functions.php#L162-L167

asharirfan:
@westonruter This PR was only about removing styles for Jetpack Infinite Scroll. Do you want to remove the theme support for infinite scroll too?

https://github.com/WordPress/twentynineteen/blob/ebf80e929d3f87014383cb279a14132282373f2a/inc/jetpack.php#L19-L24

westonruter:
I expect that the core theme would not have anything specific to Jetpack, so yes, `infinite-scroll` theme support would also be removed.

allancole:
Yup, good catch @westonruter. 

Could you or @asharirfan do a new PR that removes those remaining Jetpack references?

asharirfan:
Sure @allancole 🙌

PR #60, opened 2018-10-16
Small WPCS issues in site-branding.php

audrasjb:
Fixes some very small WPCS issues in `site-branding.php` for better readability.
Cheers,
Jb

WordPress.org username: audrasjb


allancole:
Good catch. Thanks a bunch ;-)

Issue #14, opened 2018-10-16
Add widget area(s) to the footer


kjellr:
The theme currently doesn't include any widget areas. We should add at least one to the page footer.

LittleBigThing:
I agree to have it/them in the footer. That is probably the area that can be implement the quickest while profiting from Gutenberg's new features like wide-align and columns.

Maybe interesting to check with leads of Gutenberg's phase 2 which includes "replacing widgets with blocks in registered sidebars"?

00travelgirl00:
@melchoyce Yes, that would be an option. Just in my experience, new WordPress User have a hard time to find out how to do that. Because:

1. Create the pages
2. Create an Menu
3. Find the Widget Areas
4. Find the Menu Widget 
5. Choosing the created Menu

By using the Footer Menu, the users only have to:

1. Create the pages
2. Create an Menu
3. define the position 

it is also easier to have an overview about the menus, because of the second tab in "Apperance" - "Menus"

Are there any points agains an footer menu position?

joyously:
I would like to see a normal sidebar. I really don't like web sites with no sidebars. They feel so dead-ended and static.

nielslange:
@kjellr It looks like the footer widget area issue would be solved with @fabiankaegy's PR #61, wouldn't it?

fabiankaegy:
@nielslange I believe the problem is that I haven’t gotten around to adding styles for the core widgets jet. So if someone wants to take that up they are welcome to do so. At the moment I’m working on getting #113 done.

nielslange:
@fabiankaegy Thanks for your reply. I'll take over this issue then. ;)

kjellr:
Thanks, @fabiankaegy + @nielslange for your progress on these issues!

allancole:
Closing this since the solution got merged in with #61

PR #61, opened 2018-10-16
Added two widgetareas before the siteinfo

fabiankaegy:
No styling applied jet. Only inserted the two areas. 

WordPress.com Username for contributors.md:

@fabiankaegy

### DIFF
--- a/template-parts/footer/footer-widgets.php
+++ b/template-parts/footer/footer-widgets.php
@@ -1,0 +1,20 @@
+<?php
+if ( is_active_sidebar( 'sidebar-1' ) ||
+	 is_active_sidebar( 'sidebar-2' ) ) :
+?>
+
+	<aside class="widget-area" role="complementary" aria-label="<?php esc_attr_e( 'Footer', 'twentynineteen' ); ?>">
+		<?php
+		if ( is_active_sidebar( 'sidebar-1' ) ) { ?>
+			<div class="widget-column footer-widget-1">
+				<?php dynamic_sidebar( 'sidebar-1' ); ?>
+			</div>
+		<?php }
+		if ( is_active_sidebar( 'sidebar-1' ) ) { ?>
+			<div class="widget-column footer-widget-2">
+				<?php dynamic_sidebar( 'sidebar-2' ); ?>
+			</div>
+		<?php } ?>
+	</aside><!-- .widget-area -->
+
+<?php endif; ?>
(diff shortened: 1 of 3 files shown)


fabiankaegy:
#14

carolinan:
Why does it need to be two widget areas?

Is there any difference between the two areas? If there is, I would suggest making the descriptions clearer.

One area should be able to fit any number of widgets.

template-parts/footer/footer-widgets.php
```diff
@@ -0,0 +1,20 @@
...
+?>
+
+	<aside class="widget-area" role="complementary" aria-label="<?php esc_attr_e( 'Footer', 'twentynineteen' ); ?>">
+		<?php
+		if ( is_active_sidebar( 'sidebar-1' ) ) { ?>
+			<div class="widget-column footer-widget-1">
+				<?php dynamic_sidebar( 'sidebar-1' ); ?>
+			</div>
+		<?php }
+		if ( is_active_sidebar( 'sidebar-1' ) ) { ?>
```
carolinan: This needs to be `if ( is_active_sidebar( 'sidebar-2' ) ) { ?>`

fabiankaegy:
The reason I added two widget areas was the same for having 2 in twenty seventeen. To allow having two columns. Because most of the widgets are quite narrow (perfect for actual sidebars) it would result in a lot of whitespace only allow 1 area.

carolinan:
One widget area should be able to fit any number of widgets.

The number of columns still need to be handled with css,  depending on screen size,  independent of the number of widget areas.

allancole:
This is a great start! Just needs some layout and visual styling applied to it before it gets merged in.

fabiankaegy:
Cool I’ll ad the styles tomorrow morning 👍

kjellr:
@fabiankaegy This is generally looking good. One thought I had though was that we should include theme styles for the default widgets. For instance: we should make sure the Recent Posts widget is styled just like the Recent Posts block, etc.

fabiankaegy:
okay sure agree that they should be in there. I'll wirk on that  👍

kjellr:
I just updated the branch to include baseline styles for the default widgets:

twentynineteen test__page_id 1833

I also made some slight adjustments to the `Latest Posts`, `Archives`, and `Categories` Gutenberg blocks (in the front and back end) to match up:

- Removed commas and periods, for the same i18n reasons we removed them from the main menu. 
- Added some vertical spacing between items to better tell them apart. 
- Removed the underlines.

_Before:_
screen shot 2018-10-29 at 10 05 31 am

_After:_
screen shot 2018-10-29 at 10 05 15 am

This PR could use a review. 👍

PR #68, opened 2018-10-16
Improve search container and improve typography on 404 and no-results templates.

allancole:
This fix builds on the solution presented in #38 so that margins, vertical spacing and page-title typography aligns with the rest of the theme. 

Also, I mistakenly created this new PR with a bad name when pulling down the forked code. I’ll correct for the next time ;-)


joyously:
This just says "Improve search container", but does not address the fact that it shows the first search result title (and post meta) at the top, as if it were the page title.

kjellr:
> This just says "Improve search container", but does not address the fact that it shows the first search result title (and post meta) at the top, as if it were the page title.

@joyously would you mind providing a screenshot and/or clarification of what you mean by this? Here's what I'm seeing when I test this out: 

screen shot 2018-10-17 at 2 07 28 pm

joyously:
I'm seeing this:
search-for-test

allancole:
Oh, wow! Hadn’t seen that scenario before today. Changed my search term to “featured” and was able to replicate. I’ll work out a fix shortly @joyously 👍

Issue #43, opened 2018-10-16
404 Page Needs Styling


davidakennedy:
The 404 page needs to be styled more. It looks like it's missing some wrapper styles for widths and responsiveness.

allancole:
This is addressed in #68

designsimply:
Adding a screenshot of the 404 page for reference:

![screen shot 2018-10-18 at 5 05 24 pm](https://user-images.githubusercontent.com/1119271/47189232-d5790280-d2f8-11e8-8e6b-1021674722b4.png)
<sup>Seen at http://alittletestblog.com/thispagedoesnotexist running WordPress 4.9.8 and Gutenberg 4.0 using Firefox 62.0.3 on macOS 10.13.6.</sup>

And closing since #68 has been merged.

PR #69, opened 2018-10-16
added side margins to 404

fabiankaegy:
#43 Addes page side margins 

Username: @fabiankaegy


allancole:
Just added an almost identical fix for this in #68 amongst a few other related fixes ;-).

fabiankaegy:
just saw that  😂

Issue #41, opened 2018-10-16
Accessibility-Ready Review and Checklist


davidakennedy:
Nice work overall on the accessibility of this design. My favorite parts are the nice focus states on form fields.

Here are the [requirements I tested against](https://make.wordpress.org/themes/handbook/review/accessibility/required/).

The issues below are totally fixable with a little design love. I may have missed something, so if I did, let me know in the comments. Also, please let me know if I can retest anything after PRs come in.

Anything in marked required will need to be fixed before the theme can launch. Anything marked recommended is optional for you to consider and/or fix.

**Keyboard Navigation**

Required: The focus styles look great throughout the design. There are just a few spots that need polish.

- [x] The comment `<li>`, `.comment:hover`, has a `:hover` style. However, because it's an `<li>` and not something that can receive focus, like a link, form field or button, a user can ever make the edit link appear on focus. You should be able to show the same edit link if certain links are focused on.
- [ ] Set focus styles for the audio embed block: https://cloudup.com/caSR44Mg2IA
- [x] The focus styles on posts navigation are not showing up properly: https://cloudup.com/cdp57bMMqiP

**Controls**

Pass.

**Skip Links**

Pass.

**Forms**

Required: The comment form is hidden with `display: none;`.

- [x] Forms with more than one field need to have all labels displayed visually.

**Headings**

Pass.

- [x] Recommended: Some views, like regular pages and archive pages, have more than one `<h1>`. This is okay, but not ideal because screen reader users like the clarity of one heading at level one. See _s for the ideal way to do this: https://github.com/Automattic/_s/blob/master/header.php#L31 Already mentioned in #5.

**ARIA Landmark Roles**

Pass.

- [x] Recommended: You don't need the explicit roles on the main navigation, ` role="navigation" aria-label="<?php esc_attr_e( 'Top Menu', 'twentynineteen' ); ?>` or on the `<main>` element. They can be removed.

**Content Links**

- [x] Required: Links within content must be underlined. When links appear within a larger body of block-level content, they must be clearly distinguishable from surrounding content. This would be for links in the main post body.

**Repetitive Link Text**

Pass.

**Contrasts**

Required: There are a number of colors that aren't high contrast enough to meet the requirements. Most of the colors in the design are though.

Color issues:

- [x] #919191 on #ffffff – Found in post and page meta. On search pages.
- [x] #00a0d2 on #ffffff – The hover color found on some links.

Hover colors also need to pass contrast rules.

**Images**

Pass.

[...]

PR #92, opened 2018-10-17
Remove unnecessary and redundant ARIA roles [a11y]

audrasjb:
Related: #41

>  Recommended: You don't need the explicit roles on the main navigation, role="navigation" aria-label="<?php esc_attr_e( 'Top Menu', 'twentynineteen' ); ?> or on the <main> element. They can be removed.

Cheers,
Jb

(audrasjb on w.org)

### DIFF
--- a/archive.php
+++ b/archive.php
@@ -11,7 +11,7 @@
 get_header(); ?>
 
 	<div id="primary" class="content-area">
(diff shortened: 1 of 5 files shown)


allancole:
Looks good, @audrasjb ;-)

PR #96, opened 2018-10-17
Replace the_content with the_excerpt when not in a singular post/page

### LINKED ISSUE #89: full content versus excerpts
Can we make an option to choose full content or excerpts?
Full content is not that good for SEO, and really bad user experience on search page.

audrasjb:
Fixes #89 
Indeed, when we are not on a singular post/page, we should avoid to display the full post content.


audrasjb:
I made another PR to handle this. See #244

PR #99, opened 2018-10-17
Remove unnecessary use of alt text in image ::after (CSS/SASS)

### LINKED ISSUE #65: Untranslatable text in CSS
https://github.com/WordPress/twentynineteen/blob/7b1170c700b51ccc89479a5552c97102832b2d77/style.css#L717-L727

Why would you want to do this for broken images? It would be better for the user if the alt text is shown and/or the user's setting in his browser is used for showing broken images.

Since psuedo elements on `img` tags only work when the image is not there, I think this CSS (for :before and :after) should be removed.

audrasjb:
Fixes #65 
Removes unnecessary alt text (which was even not translatable) from ::after CSS/SASS rule.

Screenshots:

Before:
capture d ecran 2018-10-17 a 04 18 50
After:
capture d ecran 2018-10-17 a 04 20 12

And also the same goes for the potential custom logo of the website (same CSS rule).
Before:
capture d ecran 2018-10-17 a 04 18 59
After:
capture d ecran 2018-10-17 a 04 19 55

Cheers,
Jb

audrasjb on w.org

### DIFF
--- a/sass/elements/_elements.scss
+++ b/sass/elements/_elements.scss
@@ -78,7 +78,7 @@ img {
 
 	&:after {
 		color: $color__text-input;
-		content: "This image is broken :-/ ( "attr(alt)" )";
+		content: " ";
 		display: block;
 		left: 50%;
 		position: absolute;
--- a/rtl.css
+++ b/rtl.css
@@ -9,5 +9,4 @@ https://codex.wordpress.org/Right-to-Left_Language_Support
 */
 body {
   direction: rtl;
-  unicode-bidi: embed;
-}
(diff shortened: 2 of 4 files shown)


sass/elements/_elements.scss
```diff
@@ -78,7 +78,7 @@ img {
 
 	&:after {
```
allancole: @audrasjb When you get a chance, please remove both the `&:before` and `&:after` pseudo elements entirely and we can get this merged in. Thanks! 👍

audrasjb:
Closed in favor of #317

PR #101, opened 2018-10-17
pointing get_template_part to correct path for content-none.php

### LINKED ISSUE #97: path to template part is wrong
https://github.com/WordPress/twentynineteen/blob/be850d11ee31434ad700db518ff8570f82e65b9c/search.php#L51
and
https://github.com/WordPress/twentynineteen/blob/be850d11ee31434ad700db518ff8570f82e65b9c/archive.php#L45

reference a template part that does not exist.
It would be better for users if this was not so nested. In order for a child theme to replace a template part, the same directory structure has to be made. Flatter is better. Also, calling one folder `post` is a little odd considering it should work for any post type.

yingles:
Updated `get_template_part` path for `content-none.php` in `archive.php` and `search.php`.

Fixes #97 

WordPress.org name: yingling017


yingles:
Second commit removes the `get_template_part( 'template-parts/post/post', 'navigation' ); ` call within `single.php`. The `post-navigation.php` file does not exist and seems to be redundant since the pages uses the `the_post_navigation` function already.

Fixes: #102 

WordPress.org name: yingling017

allancole:
Had to re-sync this with `master` and a few things were replaced with other functionality, but the fix for `search.php` checks out. Thanks for contributing 👍

Issue #105, opened 2018-10-17
min-height for the img element?


LittleBigThing:
Currently, there is a `min-height` set for the `img` element using SASS:

```
img {
    min-height: calc(3 * #{$size__spacing-unit});
}
```

I think there are some problems here:
- it is not clear for me why a min-height is needed
- it should probably not use `calc()` as it results in `calc(3 * 1rem)`in CSS instead of 3rem what is probably meant to happen
- it affects emojis as well:

<img width="624" alt="screenshot 2018-10-17 at 08 51 29" src="https://user-images.githubusercontent.com/5901923/47067979-5a75f600-d1eb-11e8-9481-1c076ce9f2d6.png">

So if we keep it, we should set it as

```
img {
    &:not(.emoji),
    &:not(.wp-smiley) {
        min-height: 3 * #{$size__spacing-unit};
    }    
}
```

<img width="623" alt="screenshot 2018-10-17 at 08 51 46" src="https://user-images.githubusercontent.com/5901923/47068173-e12ad300-d1eb-11e8-9caa-1e190975963e.png">

LittleBigThing:
`calc()` is meant to calculate a mathematical expression in CSS, but mostly useful if the calculation involves different units. For example, `calc(100% - 20px)`. The above is not really a useful expression, except when the variable is a CSS custom property.

kjellr:
Fixed via #106. The `calc()` in question was removed, so we can table that for now. 🙂

PR #106, opened 2018-10-17
Feature/min height emoji fix

fabiankaegy:
fixes issue mentioned in issue #105


fabiankaegy:
This fix introduces an error where in the admin avatar image also gets the `min-height` Banause of the specificity of the `:not()`selector. Working on a fix for that

LittleBigThing:
It should be `min-height`.

fabiankaegy:
@LittleBigThing thanks for noticing. Stupid mistake  😂

kjellr:
@fabiankaegy did you try just removing the `min-height` entirely? I'm not 100% sure it's needed.

allancole:
Yup, @fabiankaegy you can remove `min-height` entirely from the `img` selector—it’s unnecessary. 

Take that out and we can go ahead and merge this PR ;-)

PR #108, opened 2018-10-17
Let long preformatted text overflow

### LINKED ISSUE #84: content chopped off
Long lines in a `pre` tag are chopped off, probably with overflow hidden on the site-content.
It would be best to have `pre { overflow: auto }`

nielslange:
Fixes #84

<table>
<tr>
<td>Before:

84 - before
</td>
<td>After:

84 - after
</td>
</tr>
</table>


allancole:
Looks good! Thanks for contributing 👍

PR #109, opened 2018-10-17
removed period from site description

### LINKED ISSUE #103: Remove Period from Site Description
Adding a period on the site description within the code is a bit presumptuous of the expected content. A user may choose to end the site description with alternate punctuation or even an emoticon. ❤️❤️

end-with-question-mark
end-with-emoticon

However, more importantly, many languages do not use punctuation to designate the end of a sentence.

Proposal here is to remove the statically coded period from the template and allow users to choose which punctuation, if any, they want to use.

enodekciw:
Closes #103


kjellr:
This seems reasonable, and the code looks good. Thanks, @enodekciw!

Issue #117, opened 2018-10-17
Confusing display of recent posts block - grid view


fabiankaegy:
In the grid view the comma separation of the menu starts to fall apart.

<img width="876" alt="bildschirmfoto 2018-10-17 um 15 43 38" src="https://user-images.githubusercontent.com/20684594/47090645-9b89fc80-d223-11e8-8c61-7278e9121876.png">

kjellr:
Yes, this aspect hasn't been truly styled yet. Adding the `needs design` tag for now, someone should feel free to pick it up if they have time.

fabiankaegy:
<img width="849" alt="bildschirmfoto 2018-10-17 um 15 57 41" src="https://user-images.githubusercontent.com/20684594/47091563-654d7c80-d225-11e8-992a-b336765d64e5.png">
How about something like this?

fabiankaegy:
Maybe with the text aligned at the bottom of the rectangle?

kjellr:
Something like this could also keep with our text-only aesthetic:

<img width="626" alt="screen shot 2018-10-17 at 10 42 37 am" src="https://user-images.githubusercontent.com/1202812/47094531-649ef100-d1f9-11e8-8985-b249fd471dbb.png">

fabiankaegy:
I like the bold options but agree that it might be a better user experience to have the content stand out more that additional posts. With the full color background or even the back it might take too much attention away from other elements

kjellr:
Closed via #120

PR #120, opened 2018-10-17
styling for latest posts grid view

fabiankaegy:
As discussed in #117 here is an implementation of the last design mentioned by @kjellr. 
bildschirmfoto 2018-10-17 um 17 12 17


kjellr:
This looks great, thanks @fabiankaegy. One small suggestion — let's use `$color__border-input` (`#ccc`) instead, so we don't introduce a brand new color here. (May also want to rename that variable to be more generic to avoid confusion).

fabiankaegy:
On it  👍

kjellr:
LGTM. Thanks @fabiankaegy!

PR #124, opened 2018-10-17
lower overlay opacity

### LINKED ISSUE #121: Adjust opacity of filter/overlay layer for IE and Edge
The blue layer that is set over featured images has a distinct effect and it makes the theme kind of unique (among default themes). However, `mix-blend-mode` that is used to achieve this is not supported by, for example, Internet Explorer and Edge. In these browsers, a blue rectangle is displayed without any transparency: the featured image is not visible at all.

To kind of change this, the opacity value of the overlay could be lowered. This results in a similar (but different :-)) effect in browsers not supporting `mix-blend-mode`.

We should probably look for a better solution...
Also, the color of the layer should be customisable.

LittleBigThing:
Temporarily fixes #121 

Lower the opacity of the overlay filter on featured images

### DIFF
--- a/sass/site/header/_site-featured-image.scss
+++ b/sass/site/header/_site-featured-image.scss
@@ -136,7 +136,7 @@
 	.hentry:after {
 		background: $color__link;
 		mix-blend-mode: multiply;
-		opacity: 1;
+		opacity: 0.69;
 		z-index: 3;
 	}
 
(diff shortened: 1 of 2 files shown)


kjellr:
Thanks for the exploration! This was a great idea, but I think we'll need to come up with something else: lowering the opacity of the overlay can negatively effect the text contrast of the content that sits on these images.

_Before_
screen shot 2018-10-18 at 9 05 26 am

_After_
screen shot 2018-10-18 at 9 05 10 am

I'm going to close this one for that reason, but we can discuss other options in the main thread.

Issue #110, opened 2018-10-17
Missing site description 


dartiss:
The site description is not appearing on individual post or pages, only on the main archive list. However, the title and dash is appearing. 

Here it is being displayed - https://pento.net

Here is it now missing - https://pento.net/2018/10/04/wordpress-5-0-needs-you/

enodekciw:
As far as I understand, hiding it on individual post pages is a design decision ([see this](https://github.com/WordPress/twentynineteen/blob/f0b3e8cdb908aa3cb0a7171dde3c749632c51612/template-parts/header/site-branding.php#L11)).

Strange mdash behaviour should be resolved if #19 is merged

dartiss:
That's... odd 🤔Why would you want a reduced branding on the vast majority of the site?

At the very least I think this should be an option in the customizer

patilswapnilv:
I agree with @dartiss . We can have an option in the customizer. Having description would be required in number of instances.

nielslange:
+1 for @dartiss explanation. This might lead to some dirty hacks to manually add the tagline on other pages.

nielslange:
I created #142 to remove the trailing dash from the site title in case the tagline is empty. This fix is independent of the pages. The trailing dash will also disappear on homepage and blog posts page if no tagline had been provided by the user.

designsimply:
Related: #17 (duplicate even?)

nielslange:
> Related: #17 (duplicate even?)

Yes, looks pretty duplicated to me, @designsimply. I just closed #142 as #129 is a more elegant solution.

PR #129, opened 2018-10-17
display site-description on all pages and move mdash to css

enodekciw:
more complete solution to #17 
as suggested in #110, now shows .site-description on all the pages, not only `is_home()` or `is_front_page()`.


designsimply:
Thank you for contributing! Very cool.

I noticed #142 also mentions removing the emdash. Can you coordinate with @nielslange about that part? 🙂

nielslange:
@enodekciw I think your solution is a bit more elegant. Thus, I'm going to close #142. 😉

Issue #91, opened 2018-10-17
gallery shortcode columns not working


joyously:
There are styles for gallery shortcode columns, but it is incorrect.
For 1 column, the images are centered, but the captions are left. There is no space between images with no captions.
For 2 columns, 3 are shown.
For 3 columns, 2 are shown.
For 4 through 9 columns, it looks almost like 1 column, but the images are very far left on 4, a little more right on 5, more right on 6, etc., but all have just 1 image instead of 4, 5, 6, etc.

briannaorg:
I'm working on this one.

kjellr:
@joyously would you mind providing a screenshot here so we can get a better sense of the issue? Thanks!

briannaorg:
I've submitted a fix for this - https://github.com/WordPress/twentynineteen/pull/130

joyously:
Here is what it looked like on my test site
![gallery-test](https://user-images.githubusercontent.com/7957680/47105414-75576300-d209-11e8-8686-2afe79bfa857.jpg)

kjellr:
Fixed via https://github.com/WordPress/twentynineteen/commit/918a410b23d8342a0740f5e17506b8bea770ef75

kjellr:
@allancole Now that I'm looking at this again, it's not looking fixed to me:

<img width="1251" alt="screen shot 2018-10-24 at 10 06 52 pm" src="https://user-images.githubusercontent.com/1202812/47471660-4d539b00-d7d9-11e8-9112-c745031f039f.png">

(Using the [theme unit test](https://github.com/WPTRT/theme-unit-test) sample posts)

kjellr:
Closing this one in favor of the newer issue, #531.

Issue #59, opened 2018-10-16
Layout of core/gallery block on small screens


fabiankaegy:
<img width="504" alt="bildschirmfoto 2018-10-16 um 21 45 46" src="https://user-images.githubusercontent.com/20684594/47044735-42718880-d191-11e8-9b1b-b797f5be2ddf.png">

The alignment is broken and the last item is displayed onto of the other items and the element before it.

Issue #58, opened 2018-10-16
Issues with the layout core/gallery block because margin top on first element


fabiankaegy:
<img width="545" alt="bildschirmfoto 2018-10-16 um 21 45 42" src="https://user-images.githubusercontent.com/20684594/47043068-fae8fd80-d18c-11e8-8fa0-ee9abe01fe49.png">
<img width="1097" alt="bildschirmfoto 2018-10-16 um 22 15 54" src="https://user-images.githubusercontent.com/20684594/47044664-135b1700-d191-11e8-91cb-bf0e548b9759.png">

fabiankaegy:
The visual jump on the first image in the gallery is caused by line 30 in `sass/blocks/_blocks.scss`.
```sass 
.entry-content > * > *:first-child, 
.entry-summary > * > *:first-child {
    margin-top: 0;
}
```

fabiankaegy:
It also applies to `:last-child` `margin-bottom: 0;`

fabiankaegy:
<img width="999" alt="bildschirmfoto 2018-10-16 um 22 39 12" src="https://user-images.githubusercontent.com/20684594/47045948-5b2f6d80-d194-11e8-8f0f-88526140a8de.png">

joyously:
There was discussion in #core-editor about how to fix margin collapsing in general.

fabiankaegy:
This also seems to be fixed in Gutenberg 4.0 
<img width="504" alt="bildschirmfoto 2018-10-18 um 09 18 32" src="https://user-images.githubusercontent.com/20684594/47138711-709eb780-d2ba-11e8-82c4-c3e79850ccb8.png">

PR #130, opened 2018-10-17
Corrected column math in galleries - issue #91

briannaorg:
contributor: technosiren

1) If columns are being displayed inside .entry-content, as in the gallery use case, then vw is not the correct unit of measure for the column map because the entire viewport isn't available. I switched the columns map to use percentages instead, so it's sized to its container. 

2) As pretty as the column map is in _columns.scss, the math doesn't actually work. 

A single-column gallery is not equivalent to a single column width - it's equal to 12 column widths. I adjusted the math in the loop inside _galleries.scss to accommodate this, and update the map key in _columns.scss to accommodate decimal values for the actual widths required. 

While testing, I discovered that the values returned for the 7 and 9 column galleries weren't mappable due to decimal places. So I ended up eliminating the columns map entirely from _galleries.scss and left it as the original map with viewport units, not currently in use.

3) I included the gutters and inherited text alignment from the .gallery-item container so every iteration of gallery columns 1-9 is sized appropriately for its container.


joyously:
How do I test this with the other CSS changes?

joyously:
Yeah, so I'd have to do that for the other changes too, to test them together...

briannaorg:
> Yeah, so I'd have to do that for the other changes too, to test them together...

Not if you first checkout my branch with the fix into your working directory - then you're testing only the code in that branch.

joyously:
I know how to test one at a time. I'm trying to test them together.

briannaorg:
Sorry, maybe posting links to the issues you're trying to test with would help - I can either merge my fixes with those or we can ask @kjellr for testing guidance.

kjellr:
Which other changes are you trying to test this with, @joyously? In general, we tend to test changes one at a time, within individual branches.

joyously:
I think it was the similar issues: #59 and #58

briannaorg:
I believe those are coming from Gutenberg, however I have not yet been able to duplicate them.

briannaorg:
@allancole @kjellr this one is finally conflict-free and ready to go.

PR #132, opened 2018-10-17
Remove unnecessary display property from .alignleft and .alignright

### LINKED ISSUE #71: unrelated property in alignleft, alignright
https://github.com/WordPress/twentynineteen/blob/7b1170c700b51ccc89479a5552c97102832b2d77/style.css#L1205-L1215

Specifying `display:inline` has the potential to interfere with other styles, as these classes can apply to anything, especially a lot of legacy content. It is unnecessary and would be better if removed.

kjellr:
Removes `display: inline;` from `.alignleft` and `.alignright` in `_alignments.scss`. This seemed to have no negative effects on the front end. 

Fixes #71

### DIFF
--- a/sass/modules/_alignments.scss
+++ b/sass/modules/_alignments.scss
@@ -1,11 +1,9 @@
 .alignleft {
-	display: inline;
 	float: left;
 	margin-right: $size__spacing-unit;
 }
 
 .alignright {
-	display: inline;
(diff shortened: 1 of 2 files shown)


allancole:
Yup, just tested this and it holds up on both the frontend and backend.

PR #133, opened 2018-10-17
Added custom header support for the Theme

mmaumio:
It added the checkbox to display the title/tagline as well.
Allowed users to upload a header image from the customer.
Allowed the option to change the Site Title and Description.
As It was present on the other Bundled themes.

WordPress.org Username: mmaumio


inc/custom-header.php
```diff
@@ -0,0 +1,74 @@
...
+ * @package twentynineteen
+ */
+
+/**
+ * Set up the WordPress core custom header feature.
+ *
+ * @uses twentynineteen_header_style()
+ */
+function twentynineteen_custom_header_setup() {
+	add_theme_support( 'custom-header', apply_filters( 'twentynineteen_custom_header_args', array(
```
celloexpressions: Add the following line to enable header video support. See #243:

`		'video'                  => true,`

inc/custom-header.php
```diff
@@ -0,0 +1,74 @@
+<?php
+/**
+ * Sample implementation of the Custom Header feature
+ * http://codex.wordpress.org/Custom_Headers
+ *
+ * You can add an optional custom header image to header.php like so ...
```
celloexpressions: This comment should probably be removed

inc/custom-header.php
```diff
@@ -0,0 +1,74 @@
+<?php
+/**
+ * Sample implementation of the Custom Header feature
```
celloexpressions: Revise this comment

inc/custom-header.php
```diff
@@ -0,0 +1,74 @@
...
+add_action( 'after_setup_theme', 'twentynineteen_custom_header_setup' );
+
+if ( ! function_exists( 'twentynineteen_header_style' ) ) :
+	/**
+	 * Styles the header image and text displayed on the blog.
+	 *
+	 * @see twentynineteen_custom_header_setup().
+	 */
+	function twentynineteen_header_style() {
+		$header_text_color = get_header_textcolor();
```
celloexpressions: I recommend turning off support for the header text color. An overlay is provided to ensure readability. See also the discussions for Twenty Seventeen.

If header text color is a feature, it should be coordinated with the other color customization options instead of being implemented here.

inc/template-functions.php
```diff
@@ -22,6 +22,11 @@ function twentynineteen_body_classes( $classes ) {
 		$classes[] = 'hfeed';
 	}
 
+	// Add a class if there is a custom header.
+	if ( has_header_image() ) {
```
celloexpressions: Revise to:
`if ( has_header_image() || has_header_video() ) {`

inc/template-functions.php
```diff
@@ -22,6 +22,11 @@ function twentynineteen_body_classes( $classes ) {
...
+		$classes[] = 'has-header-image';
```
celloexpressions: Revise to:
`$classes[] = 'has-header-media';`

js/customizer.js
```diff
@@ -7,5 +7,37 @@
  */
 
 ( function( $ ) {
+
```
celloexpressions: I recommend leaving any color customization or other changes to this file out of this PR, and coordinating those with the proposed color customization options.

style.css
```diff
@@ -1259,14 +1259,36 @@ body.page .main-navigation {
...
-  padding: 1em;
+
+.custom-header {
+    position: relative;
+    padding: 1rem;
+    display: block;
+    height: auto;
+}
+
+.has-header-image .custom-header .wp-custom-header img {
```
celloexpressions: Add the selector for header video.

style.css
```diff
@@ -1259,14 +1259,36 @@ body.page .main-navigation {
 /*--------------------------------------------------------------
 ## Header
 --------------------------------------------------------------*/
```
celloexpressions: Add this first: 
```
.wp-custom-header iframe,
.wp-custom-header img,
.wp-custom-header video {
	display: block;
	height: auto;
	max-width: 100%;
}
```

style.css
celloexpressions: Which is probably:

`.has-header-media .custom-header .wp-custom-header video`

joyously:
I think this is good to have, regardless of the editor used.
But does it consider what image to show if there is a header image _and_ a featured image?

mmaumio:
So, are we keeping the custom header support off for now?
If not, please let me know if there's any enhancement I can do on the PR to get accepted.

kjellr:
> So, are we keeping the custom header support off for now?

Yes  — as per @allancole's notes on focusing on Gutenberg, let's leave this out for now. I'll go ahead and close this. 

Thank you for the PR in any case!

Issue #87, opened 2018-10-16
Centered captioned images are not centered


joyously:
The caption is centered under the image, but the image is not centered in the content.

khleomix:
@kjellr I'll take this up if no one's working on it.

khleomix:
caption is not centered in Gutenberg-inserted image.  `.entry-content .wp-block-image figcaption` is overriding `.wp-block-image figcaption` 
@kjellr should this be a separate issue? If so, I'll create one for it if it's not yet addressed.

joyously:
I wrote one about having duplicate CSS and too specific selectors.

khleomix:
@joyously added my comment to your issue for more reference. Thanks.

Issue #77, opened 2018-10-16
duplicate, more specific CSS


joyously:
There is a `Blocks` section in the CSS that has the `entry-content` class at the front of all selectors. Some rules are duplicates of earlier rules, but more specific. Can these rules work without the `entry-content` class or do they need to be that specific?

khleomix:
this is affecting the caption for Gutenberg-inserted images:
caption is not centered in Gutenberg-inserted image. `.entry-content .wp-block-image figcaption` is overriding `.wp-block-image figcaption`

kjellr:
The `.entry-content` class is required to make sure the blocks that have no classnames inherit styles correctly (`p`, `h1`, etc.). Outside of those, there may be some specific cases that we can address, however I'd suggest we address those in individual issues on Trac at this point. 

I'm going to close this issue for now, but feel free to open up a Trac issue for specific enhancements/code cleanup in this area.

PR #134, opened 2018-10-17
Fix for: Centered captioned images are not centered - issue #87

khleomix:
contributor: @khleomix

The image centering issue is not caused by CSS but rather by the injected hard-coded style added by the caption shortcode. Add a function to address that. 

For the style of the caption, it should be the same as the Gutenberg inserted one now. This might change if #77 can be addressed.


joyously:
> The image centering issue is not caused by CSS but rather by the injected hard-coded style added by Gutenberg.

This is not the case, because I am not using Gutenberg.

joyously:
If a user chose a size for the image, and core is adding that size to the style tag, the user will get an image the size he chose. If you remove the size with this filter, what will he get? This issue is about centering, not about changing the size.

khleomix:
@joyously sorry, my mistake. The injected style is added by the wp-caption shortcode

khleomix:
@joyously I've tested with and without Gutenberg and with different image sizes and tested with left and right aligned images and in mobile and tablet without the filter and everything checks out. I've checked out themes and created themes that disable the filter as well without any issues. I could not find any other cause for the centering problem aside from that filter - it might be caused by CSS and might be related to #77. I'll check this further tomorrow when I wake up.

khleomix:
Furthermore, the Gutenberg image block outputs the same code but without the injected style and that seems to work pretty well.

kjellr:
The issue doesn't seem to be fixed here for me. I'm also seeing a caption alignment bug that I thought we fixed in #145. 🤔 

**Front End:**
screen shot 2018-10-27 at 12 58 51 pm

**Editor:**
screen shot 2018-10-27 at 12 58 08 pm

The margin for the `align center` class appears to be overridden by [this global rule](https://github.com/WordPress/twentynineteen/blob/8d47b1b791436bae516a1668828d35f1374d3df5/sass/blocks/_blocks.scss#L6).

kjellr:
I'm going to close this one, since #380 took care of the centering issue. I'll borrow code from this to sync up the caption styles in a separate PR.  

Thanks everyone!

Issue #90, opened 2018-10-16
<big> is too big


joyously:
The `<big>` tag is too big, and is one size only, so nested tags are not bigger. Note how the spacing is all awry.

![big-tag-too-big](https://user-images.githubusercontent.com/7957680/47053836-72f6f980-d174-11e8-96f5-98f0fbab5962.jpg)

nielslange:
@kjellr or @allancole Can one of your close all merged issues, to retain an overview of the open issues? If needed, I'm happy to go through all issues and provide a list of merged issues. 😉

kjellr:
@nielslange Thanks! I'll do a quick scrub of these today. 

This particular one is closed via #137.

nielslange:
Thank, @kjellr!

PR #137, opened 2018-10-17
changed size of <big> element to use the lg font size

fabiankaegy:
bildschirmfoto 2018-10-18 um 00 08 19


fabiankaegy:
fixes issue #90

PR #138, opened 2018-10-17
Add Gulp tasks for compiling scss

dereksmart:
This adds a very basic Gulp process for compiling scss.

It currently only compiles to `style.css`, and does nothing for RTL, or the editor-style.css, etc. 

It can be expanded to create individual .css files in the `sass/` directory if needed. This is just a starting point for whatever is actually needed. 
 
To test (should probably be added to documentation somewhere): 
- Make sure you have `npm` [installed](https://www.npmjs.com/get-npm) 
- `npm install` to build the packages
- `gulp sass` (should build the new style.css)
- `gulp sass:watch` (should automatically build new css file on changes to any scss file in `/sass`)

cc @allancole

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .DS_Store
+node_modules/
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,0 +1,33 @@
+// Imports
+const gulp = require( 'gulp' ),
+    banner = require( 'gulp-banner' ),
+    sass = require( 'gulp-sass' ),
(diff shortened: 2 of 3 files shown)


ntwb:
WordPress uses Grunt across all of its projects, that should also be the case here.

It helps to lessen the learning curve by keeping the tooling somewhat consistenet accross projects.

Note: WordPress as part of 5.0 will use webpack, and as part of 5.1 will drop Grunt in favour of webpack and npm scripts.

ntwb:
Let's go with _npm scripts_ per #169 instead of Gulp or Grunt

PR #140, opened 2018-10-18
Add print styles

### LINKED ISSUE #88: Where are the print styles?
There should be at least a little bit of print styles to hide sidebars and menus and other interactive elements.

nielslange:
Fixes #88

[PDF before](https://github.com/WordPress/twentynineteen/files/2489779/88.-.before.pdf) | [PDF after](https://github.com/WordPress/twentynineteen/files/2489778/88.-.after.pdf)

Note: This is a first draft that should be extended once #61 had been merged.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -119,6 +119,8 @@ function twentynineteen_scripts() {
 
 	wp_enqueue_script( 'twentynineteen-skip-link-focus-fix', get_template_directory_uri() . '/js/skip-link-focus-fix.js', array(), '20151215', true );
 
+	wp_enqueue_style( 'twentynineteen-print-style', get_template_directory_uri() . '/print.css', array(), filemtime(get_template_directory() . '/print.css'), 'print' );
+
 	if ( is_singular() && twentynineteen_can_show_post_thumbnail() ) {
 		wp_add_inline_style( 'twentynineteen-style', twentynineteen_header_featured_image_css() );
 	}
--- a/functions.php
+++ b/functions.php
@@ -119,7 +119,7 @@ function twentynineteen_scripts() {
 
 	wp_enqueue_script( 'twentynineteen-skip-link-focus-fix', get_template_directory_uri() . '/js/skip-link-focus-fix.js', array(), '20151215', true );
 
-	wp_enqueue_style( 'twentynineteen-print-style', get_template_directory_uri() . '/print.css', array(), filemtime(get_template_directory() . '/print.css'), 'print' );
+	wp_enqueue_style( 'twentynineteen-print-style', get_template_directory_uri() . '/print.css', array(), wp_get_theme()->get( 'Version' ), 'print' );
 
 	if ( is_singular() && twentynineteen_can_show_post_thumbnail() ) {
 		wp_add_inline_style( 'twentynineteen-style', twentynineteen_header_featured_image_css() );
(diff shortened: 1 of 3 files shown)


functions.php
```diff
@@ -119,6 +119,8 @@ function twentynineteen_scripts() {
 
 	wp_enqueue_script( 'twentynineteen-skip-link-focus-fix', get_template_directory_uri() . '/js/skip-link-focus-fix.js', array(), '20151215', true );
 
+	wp_enqueue_style( 'twentynineteen-print-style', get_template_directory_uri() . '/print.css', array(), filemtime(get_template_directory() . '/print.css'), 'print' );
```
joyously: This looks odd. Is it correct?

print.css
```diff
@@ -0,0 +1,87 @@
...
+# Page breaks
+# Links
+# Visibility
+--------------------------------------------------------------*/
+@media print {
+  /* Margins */
+  @page {
+    margin: 2cm;
+  }
+  .hentry, .hentry .entry-header, .site-footer .site-info {
```
joyously: Do not style `hentry`!  There is already an issue to remove styles for `hentry`.

functions.php
enodekciw: As mentioned in the docs:

> If version is set to false, a version number is automatically added equal to current installed WordPress version. If set to null, no version is added.

At the moment, main stylesheet has no specific version number supplied and falls back to `4.9.8` (current WordPress version). Is there a need to have a dynamic version number added for `print.css`?

joyously: I think it was because of my bad monitor, I didn't recognize the function (was seeing n instead of m), and when I suggested a similar way of doing version number for `_s`, it was shot down. I think they said that the file time function was not necessarily available on every server or something like that. I'd have to search for the issue to get the exact reason.

nielslange: > Is there a need to have a dynamic version number added for print.css?

@enodekciw I only added the dynamic version number due to testing purposes to avoid caching issues. 

> I think it was because of my bad monitor, I didn't recognize the function (was seeing n instead of m), and when I suggested a similar way of doing version number for _s, it was shot down. I think they said that the file time function was not necessarily available on every server or something like that. I'd have to search for the issue to get the exact reason.

@joyously As mentioned above, I primarily use `filemtime` to avoid caching issues. However, as `filemtime` is plain PHP I wonder why it should not be available on some servers. 🤔

designsimply:
Adding a side-by-side screenshot of page one from @nielslange's PDFs linked above for a visual reference:

screen shot 2018-10-19 at 12 50 13 pm
<sup>Before `print.css` is on the left and after `print.css` is on the right.</sup>

😍

functions.php
allancole: Hey @nielslange, Since it sounds like `filemtime` is only used for caching purposes, I think we can live without it. I think the version number should be synced with the theme’s version number. Remove that when you can and we can go ahead and merge this ;-)

nielslange: @allancole I just updated my PR and replaced `filemtime(get_template_directory() . '/print.css')` with `wp_get_theme()->get( 'Version' )`, to show the themes verion number instead.

kjellr:
I just pushed a minor update to this PR:

- Moved each selector onto their own lines ([reference](https://make.wordpress.org/core/handbook/best-practices/coding-standards/css/#structure))
- Set `main-navigation` + `social-navigation` to hide, as per the suggestion in the original ticket. 

I'm going to go ahead and merge this one in for now. Thanks again, @nielslange!

nielslange:
Thanks for merging this PR, @kjellr!

PR #141, opened 2018-10-18
Scale down our largest two font sizes on mobile.

kjellr:
This PR knocks instances of `$font__size-xxl` and `$font__size-xl` down one smaller on the type scale when we're on smaller screens. This helps preserve a reasonable character count when we're dealing with narrower viewports.

This _does_ mean that on mobile, `h2`s and `h3`s are the same type size, but it doesn't bother me because `h2`s will still have the short divider line above them to signify a greater hierarchy.

_Screenshots:_

artboard

artboard copy

artboard copy 3

artboard copy 2


allancole:
Mostly looks great! 

I noticed an issue with a right-floated blockquote.

image

Also may need to adjust the selectors for medium, large, and XL font size options in Gutenberg:

image

Otherwise, looks good everywhere else 👍

kjellr:
The right floated blockquote issue is a weird one, and I've started a local PR to get that working. In the meantime I've filed #200 to track it. 

#162 is a PR to add the Gutenberg text sizes. 

This particular one should be good to go.

PR #144, opened 2018-10-18
Fixed Latest Posts Page showing the Title and and meta data of the First Post.

mmaumio:
Latest Posts Page showing the Title and metadata of the First Post showing in the header hero area.
Whereas the page shows all the post below it.
Eventually, It was repeating the first post twice unless we ignore the first sticky posts for Main Blog page.

I'm not sure what exactly we're planning here.

For the time being the PR saves us from some confusion.

WordPress.org: mmaumio


allancole:
This is fixed in #68

PR #150, opened 2018-10-18
do not show featured image if 'your latest posts' is selected

### LINKED ISSUE #148: Featured image block is displayed when Homepage is set to 'Your latest post'
image

I think that #135 introduced this strange behaviour and now when I select 'Your latest posts' as a homepage view, it shows this blue block with the title of my last post.

Also, if I set my homepage to 'a static page' then this issue also appears on a page set as a posts page.

image

enodekciw:
Closes #148 

`! is_home()` check prevents featured image block from appearing on 'your latest posts' or 'posts page' (if homepage set to 'a static page')


enodekciw:
Fixed by #68

Issue #81, opened 2018-10-16
Comment dates are not using Settings date format


joyously:
The dates on comments are shown in relative times, which is not a core option for the date format. This is plugin territory. The date format chosen in Settings should be used.

allancole:
Closing this since a fix was merged in #151

PR #151, opened 2018-10-18
Display comment times with default functions

briannaorg:
Fixes issue #81.

Replace the lovely human_time_diff() with the get_comment_date() and
get_comment_time() functions so that a user’s chosen time display
settings are used in comments display.

If we end up building a project plugin for this (per the results of our dev tooling discussions) it would be great to add a setting in the Customizer that allows “human readable” time displays, to take advantage of the existing functions in template-tags.php.

WordPress.org username: technosiren


enodekciw:
Maybe remove `twentynineteen_human_time_diff()`completely? If it's not used anymore, it shouldn't be in `inc/template-tags.php` either.

kjellr:
> Maybe remove twentynineteen_human_time_diff()completely? If it's not used anymore, it shouldn't be in inc/template-tags.php either.

Agreed. @briannaorg — once you remove that function, this'll be good to go. 👍 

Thanks!

allancole:
Looks good! 👍Thanks @briannaorg

PR #153, opened 2018-10-18
featured image overlay pointer-events set to none

### LINKED ISSUE #152: featured image overlays prevent anchor from be clicked
Steps to reproduce:
- Write a post and set a featured image
- Visit any of the archive pages (blog index, category archive, etc)
- Try clicking on a featured image to go to a specific post

This won't work, because color overlays are "on top" of the anchor and that prevents anchor from being clicked.

enodekciw:
Closes #152


kjellr:
🚢 Thanks for the fix, @enodekciw!

Issue #149, opened 2018-10-18
Primary color set in Customizer to affect all elements. 


fabiankaegy:
I think all the elements that currently use the color `#0073aa` (Blue) should be changed using the Customizer.  

### Frontend: 
- [x] Featured Image
- [x] Post Thumbnail
- [x] Menu Item
- [ ] Menu Item Hover
- [x] Link Color ? 
- [ ] Link Color Hover
- [x] Button Color 
- [x] Pullquote left Border 
- [x] Post Navigation - Post Title 

### Customizer: 
- [x] Featured Image
- [x] Post Thumbnail
- [x] Menu Item
- [ ] Menu Item Hover
- [x] Link Color ? 
- [ ] Link Color Hover
- [x] Button Color 
- [x] Pullquote left Border 
- [x] Post Navigation - Post Title 

### Editor: 
- [ ] Link Color ? 
- [ ] Link Color Hover
- [ ] Button Color 
- [ ] Pullquote left Border 

This is supposed to extend the work of #113

fabiankaegy:
If that's ok I can take the issue and work on that. 

### Question
@kjellr there are some areas, like for example the link / menu Item color that are set using the sass function `darken()`. How should we go about setting those via plain css? Should we add a php function to mirror the behaviour of `darken()`?

fabiankaegy:
<img width="281" alt="bildschirmfoto 2018-10-18 um 14 38 08" src="https://user-images.githubusercontent.com/20684594/47154756-79f04a00-d2e3-11e8-8df8-f5a6c17aa486.png">
And I believe there either should be some warning message like in Gutenberg when the color doesn't have enough contrast to other elements, or some functionality that for example determines the font color on a button automatically based on the new background color.

killermann:
I just forked the theme and started working on exactly this, before checking the open issues. Let me know if you want help testing, @fabiankaegy.

fabiankaegy:
@killermann this is my current state: https://github.com/fabiankaegy/twentynineteen/tree/feature/primary-color-customizer I still need to add the editor styles and the functionality to warn users if there isn't sufficient contrast. Do you want to take up one of these things?

fabiankaegy:
And I'll open the pull request now so you can just have a look here. I'm happy for all the feedback I can get

fabiankaegy:
@allancole would you just add separate controls for the primary and hover color or for each element separately?

designsimply:
[...]

PR #154, opened 2018-10-18
Feature/customizer primary color

fabiankaegy:
Work in progress of #149


fabiankaegy:
@killermann this is the current state

designsimply:
@fabiankaegy thank you for working on this! I noticed #113 looks like similar work and was wondering if you'd like to keep #113 open or consolidate to #154? I know it's a best practice to use smaller, incremental PRs in general but wanted to ask you for this case.

fabiankaegy:
@designsimply I first worked on #113 and because @kjellr said that customizing the primary color of all elements should be done in a separate PR I opened this one. But If we are committing to this we can close #113 all together because this is based on the work there.

designsimply:
Ace. I vote for consolidating to #154 then! 😊

kjellr:
> I first worked on #113 and because @kjellr said that customizing the primary color of all elements should be done in a separate PR I opened this one. But If we are committing to this we can close #113 all together because this is based on the work there.

That sounds great — I definitely think it makes sense to commit to customizing all elements, so it makes sense to consolidate these. Thanks all!

Issue #94, opened 2018-10-17
Enable AMP Compatibility


amedina:
**Feature Request**
Build the theme with opt-in AMP compatibility. This way, users could start with a baseline AMP-compatible 2019 theme and build AMP-compatible sites on top of it. Would love to help with this.

westonruter:
For reference…

Example for how to add support to _s: https://github.com/Automattic/_s/pull/1286
And WP Rig has opt-in support as well: https://github.com/wprig/wprig#amp-ready

aaronjorbin:
I don't think default themes should contain code designed for a 3rd parties plugin.

westonruter:
@aaronjorbin Normally I'd agree, but it's a plugin that I work on so I make an exception 😉But Matt did suggest once that the next default theme could be canonical AMP.

Nevertheless, the main problem I see is the hard-coded use of the `is_amp_endpoint()` function in #155. What if instead of that there was a theme-specific filter which any plugin could then hook into to indicate that AMP is the target? For example:

```php
/**
 * Check if AMP plugin is enabled and if AMP endpoint.
 */
function twentynineteen_is_amp() {
	return apply_filters( 'twentynineteen_is_amp', false );
}
```

Then the AMP plugin or any other plugin that adds AMP compatibility (or child theme) would just have to do the following for Twenty Nineteen:

```php
add_theme_support( 'amp' );
add_filter( 'twentynineteen_is_amp', 'is_amp_endpoint' );
```

In this way the theme would not be including code that is specific to a 3rd-party plugin. But at the same time, it would facilitate the theme to opt-in to be used in AMPHTML, similar to how WordPress started adding support for HTML5 in addition to XHTML.

aaronjorbin:
I do think the theme should do everything in its power to make it easy for plugins to work.  I don't see where `twentynineteen_is_amp` is being used in #155 right now.  Is it not possible for the amp plugin to detect that twentynineteen is the theme and to handle everything it needs in the plugin?

westonruter:
Yeah, #155 is somewhat premature at the moment. See the pull request for _s which has examples for how this function would be used: https://github.com/Automattic/_s/pull/1286/files?w=1

Most of it boils down to not enqueueing JS when when in AMP and to instead output AMP alternatives to do things like toggling the nav menu.

[...]

PR #155, opened 2018-10-18
AMP compatibility

scottblackburn:
This is just a start on #94 , I will keep working on this PR as the theme develops to check the theme is AMP compatible. These changes just resolve some AMP validation issues popping up in the validation checker.

- Adds `twentynineteen_amp_enabled()` function to check whether response is being served as AMP and plugin is enabled.
- Don't enqueue `skip-link-focus-fix.js` on AMP responses since it is not allowed in AMP.
- Conditional added around viewport meta tag as `minimum-scale=1` is required in AMP but probably not desirable in non-AMP cases.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -117,7 +117,10 @@ add_action( 'after_setup_theme', 'twentynineteen_content_width', 0 );
 function twentynineteen_scripts() {
 	wp_enqueue_style( 'twentynineteen-style', get_stylesheet_uri() );
 
-	wp_enqueue_script( 'twentynineteen-skip-link-focus-fix', get_template_directory_uri() . '/js/skip-link-focus-fix.js', array(), '20151215', true );
+	// Don't include js on AMP endpoints.
+	if ( ! is_amp_endpoint ) {
+		wp_enqueue_script( 'twentynineteen-skip-link-focus-fix', get_template_directory_uri() . '/js/skip-link-focus-fix.js', array(), '20151215', true );
+	}
 
 	if ( is_singular() && twentynineteen_can_show_post_thumbnail() ) {
 		wp_add_inline_style( 'twentynineteen-style', twentynineteen_header_featured_image_css() );
--- a/functions.php
+++ b/functions.php
@@ -98,9 +98,9 @@ add_action( 'after_setup_theme', 'twentynineteen_setup' );
 
 
 /**
- * Set AMP flag.
+ * Check if AMP plugin is enabled and if AMP endpoint.
  */
-function is_amp_enabled() {
+function twentynineteen_amp_enabled() {
 	return function_exists( 'is_amp_endpoint' ) && is_amp_endpoint();
 }
 
@@ -126,7 +126,7 @@ function twentynineteen_scripts() {
 	wp_enqueue_style( 'twentynineteen-style', get_stylesheet_uri() );
 
 	// Don't include js on AMP endpoints.
-	if ( ! is_amp_enabled ) {
+	if ( ! twentynineteen_amp_enabled ) {
 		wp_enqueue_script( 'twentynineteen-skip-link-focus-fix', get_template_directory_uri() . '/js/skip-link-focus-fix.js', array(), '20151215', true );
 	}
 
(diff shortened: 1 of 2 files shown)


functions.php
```diff
@@ -117,7 +125,10 @@ function twentynineteen_content_width() {
 function twentynineteen_scripts() {
 	wp_enqueue_style( 'twentynineteen-style', get_stylesheet_uri() );
 
-	wp_enqueue_script( 'twentynineteen-skip-link-focus-fix', get_template_directory_uri() . '/js/skip-link-focus-fix.js', array(), '20151215', true );
+	// Don't include js on AMP endpoints.
+	if ( ! twentynineteen_amp_enabled ) {
+		wp_enqueue_script( 'twentynineteen-skip-link-focus-fix', get_template_directory_uri() . '/js/skip-link-focus-fix.js', array(), '20151215', true );
+	}
```
westonruter: This will need to change as of #47.

functions.php
```diff
@@ -117,7 +125,10 @@ function twentynineteen_content_width() {
 function twentynineteen_scripts() {
 	wp_enqueue_style( 'twentynineteen-style', get_stylesheet_uri() );
 
-	wp_enqueue_script( 'twentynineteen-skip-link-focus-fix', get_template_directory_uri() . '/js/skip-link-focus-fix.js', array(), '20151215', true );
+	// Don't include js on AMP endpoints.
+	if ( ! twentynineteen_amp_enabled ) {
```
westonruter: Missing parens on the function call.

header.php
```diff
@@ -14,7 +14,11 @@
 <html <?php language_attributes(); ?>>
 <head>
 	<meta charset="<?php bloginfo( 'charset' ); ?>" />
+	<?php if ( ! twentynineteen_amp_enabled() ) { ?>
 	<meta name="viewport" content="width=device-width, initial-scale=1" />
+	<?php } else { ?>
+	<meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
```
westonruter: Why not just have this as the default?

functions.php
```diff
@@ -96,6 +96,14 @@ function twentynineteen_setup() {
 endif;
 add_action( 'after_setup_theme', 'twentynineteen_setup' );
 
+
+/**
+ * Check if AMP plugin is enabled and if AMP endpoint.
+ */
+function twentynineteen_amp_enabled() {
```
westonruter: In WP Rig this would be called `twentynineteen_is_amp()`. I think that is better than `...amp_enabled` because your site may have AMP enabled even though it is not being served at the moment, as in the case of paired mode or if you opt-out of AMP for a given post. 

https://github.com/wprig/wprig/blob/d8d9cca76027b3057d4df2ce7cc3bd3101172281/dev/inc/template-tags.php#L10-L20

functions.php
```diff
@@ -129,6 +135,20 @@ function twentynineteen_scripts() {
...
 add_action( 'wp_enqueue_scripts', 'twentynineteen_scripts' );
 
+/**
+ * Fix skip link focus in IE11.
+ *
+ * This does not enqueue the script because it is tiny and because it is only for IE11,
+ * thus it does not warrant having an entire dedicated blocking script being loaded.
+ *
+ */
+function _twentynineteen_skip_link_focus_fix() {
```
westonruter: This change shouldn't be included _yet_. It's dependent on #47. Also, once/if that PR is merged, this function will then need the `is_amp` as seen in the WP Rig PR: https://github.com/wprig/wprig/pull/139

scottblackburn: @westonruter should we just revert back to the previous for now until #47 is merged?

westonruter: yes

header.php
```diff
@@ -14,7 +14,7 @@
 <html <?php language_attributes(); ?>>
 <head>
 	<meta charset="<?php bloginfo( 'charset' ); ?>" />
-	<meta name="viewport" content="width=device-width, initial-scale=1" />
+	<meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
```
westonruter: I've opened a separate PR just for this one change: https://github.com/WordPress/twentynineteen/pull/199

PR #156, opened 2018-10-18
Unnecessary comment in search.php

DannyCooper:
sprintf/printf is not used so this comment isn't required.


kjellr:
Thanks @DannyCooper!

PR #159, opened 2018-10-18
Update footer.php

DannyCooper:
Missing accompanying translators comment for placeholder.


kjellr:
Thanks, @DannyCooper!

PR #162, opened 2018-10-18
Add custom editor font sizes

kjellr:
Uses `add_theme_support( 'editor-font-sizes')` to allow Twenty Nineteen to customize the font size controls in Gutenberg. One caveat:

- Our current text size variables don't all output nicely rounded pixel sizes (`$font__size-md` is `36.5px` for example). We should fix that and adjust the type scale in a separate PR. Filed the issue in #164.
- `add_theme_support( 'editor-font-sizes')` only accepts pixel sizes. I've added our ugly half-pixel sizes for now, but will likely let this PR sit until we merge a PR in which font sizes work out to round numbers.

_Screenshots:_
Front end: 
screen shot 2018-10-18 at 2 26 18 pm

Editor: 
screen shot 2018-10-18 at 2 45 32 pm


kjellr:
> `add_theme_support( 'editor-font-sizes')` only accepts pixel sizes. I've added our ugly half-pixel sizes for now, but will likely let this PR sit until we merge a PR in which font sizes work out to round numbers.

After letting this sit for a few weeks, I've changed my mind. Getting the font sizes to resolve to whole pixels hasn't happened yet, but this PR is ready to go today. We might as well merge it in and update the font sizes when/if we end up revising the global font sizes before release.

cc @allancole for a review when you have a moment.

PR #163, opened 2018-10-18
Add a simple build tool to build and watch SASS files.

blowery:
This is an alternative to #138.

Testing Instructions:
* Make sure you have node installed
* `npm install`
* To build CSS from SASS: `npm run build`
* To watch for changes to SASS and rebuild when any occur: `npm run watch`

I wasn't sure how the style-editor and rtl css files were built, but this script could be adapted to automatically build those as well.

My configuration for node-sass or postcss may be slightly different from what was previously used. Some trailing `}` are now on the same line as the final statement..

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .DS_Store
+node_modules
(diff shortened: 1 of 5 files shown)


blowery:
cc @allancole

allancole:
See #169 for continued discussion…

PR #167, opened 2018-10-18
Fix linting errors causing the Travis build to fail

brandonpayton:
This PR fixes the current PHP and JS linting errors that are causing our Travis build to fail. It also adds an `.editorconfig` file that will help many editors automatically avoid introducing whitespace-related errors.

This is a combination of manual fixes and automatic fixes made by PHP Code Sniffer's `phpcbf` script.

Some the fixes involved adding comments. In those cases, I often added TODO comments so someone more familiar with the theme can explain properly.


brandonpayton:
Thanks for testing, @allancole.

I pushed another update that resolves merge conflicts introduced in the last couple of hours.

PR #169, opened 2018-10-18
Add a simple build tool to build and watch SASS files

allancole:
This builds on #163 as an alternative to #138.

Testing Instructions:

- Make sure you have node installed
- `npm install`
- To build CSS from SASS: `npm run build`
- To watch for changes to SASS and rebuild CSS when any occur: `npm run watch`


package.json
```diff
@@ -0,0 +1,26 @@
...
+  "bugs": {
+    "url": "https://github.com/WordPress/twentynineteen/issues"
+  },
+  "homepage": "https://github.com/WordPress/twentynineteen#readme",
+  "devDependencies": {
+    "autoprefixer": "^9.1.5",
+    "chokidar-cli": "^1.2.1",
+    "node-sass": "^4.9.3",
+    "postcss-cli": "^6.0.0",
+    "rtlcss": "^2.4.0"
```
blowery: you probably want to add `npm-run-all` in here using `npm i -D npm-run-all`. If `npm run build` is working right now, you're picking up `run-p` from a transitive dependency.

package.json
```diff
@@ -0,0 +1,26 @@
...
+  },
+  "browserslist": [
+    "last 2 versions",
+    "ie >= 8"
+  ],
+  "scripts": {
+    "build:style": "node-sass style.scss style.css --output-style expanded && postcss -r style.css",
+    "build:style-editor": "node-sass style-editor.scss style-editor.css --output-style expanded && postcss -r style-editor.css",
+    "build": "run-p \"build:style\" \"build:style-editor\"",
```
blowery: this can just be 
```suggestion
    "build": "run-p \"build:*\"",
```

that way when you add new stuff like `build:something-else` it'll automatically get picked up by the wildcard.

see https://github.com/mysticatea/npm-run-all/blob/HEAD/docs/run-p.md

package.json
```diff
@@ -0,0 +1,26 @@
...
+    "autoprefixer": "^9.1.5",
+    "chokidar-cli": "^1.2.1",
+    "node-sass": "^4.9.3",
+    "postcss-cli": "^6.0.0",
+    "rtlcss": "^2.4.0"
+  },
+  "browserslist": [
+    "last 2 versions",
+    "ie >= 8"
+  ],
```
ntwb: The supported browsers should be the same as core, this is documented at:
> https://make.wordpress.org/core/handbook/best-practices/browser-support/

The ideal way of supporting the above is to use the `@wordpress/browserlist-config` package:
> https://github.com/WordPress/gutenberg/tree/master/packages/browserslist-config

To do this, run `npm install @wordpress/browserslist-config --save-dev` and then use:
```json
"browserslist": [
	"extends @wordpress/browserslist-config"
]
```

package.json
```diff
@@ -0,0 +1,26 @@
+{
```
ntwb: The `package.json` file should use _tabs_ for indentation, not 2 spaces.

swissspidy: The core `.editorconfig` file specifies spaces for `package.json` though: 

https://github.com/WordPress/wordpress-develop/blob/3ba1a2b33c71a5b8dfbc6ab728bcdbff30eb1880/.editorconfig#L16-L18

swissspidy: Interesting, I didn‘t know about that limitation in npm. Figured spaces are more readable for JSON files anyway.

Since the 5.0 branch is de facto trunk right now, that‘s what I linked to.

ntwb: Right, I'm also going to backport that change for 5.0, Gutenberg switched to _tabs_ around the same time, as such I think having 5.0 and Gutenberg indentation parity would be a nice to have

allancole:
Added usage instructions to `contributors.md` and added a build process for the `print.scss` file. 

Gonna go ahead and merge this so we can all start using the same process ;-)

package.json
```diff
@@ -0,0 +1,42 @@
...
+    "plugins": [ ],
+    "map": false
+  },
+  "browserslist": [
+    "extends @wordpress/browserslist-config"
+  ],
+  "scripts": {
+    "build:style": "node-sass style.scss style.css --output-style expanded && postcss -r style.css",
+    "build:style-editor": "node-sass style-editor.scss style-editor.css --output-style expanded && postcss -r style-editor.css",
+    "build:rtl": "rtlcss style.css rtl.css",
```
ocean90: Note that `rtl.css` should only be used to overwrite the styles from the main stylesheet. Since that's not possible with rtlcss  you have change the file name to `style-rtl.css` and use `wp_style_add_data( 'twentynineteen-style', 'rtl', 'replace' );`

allancole: Ahhh, good catch! I’ll sort that out before merging.

PR #174, opened 2018-10-19
Update the `.editorconfig` file to use latest from core

### LINKED ISSUE #173: Update the `.editorconfig` file to use latest from core
The `.editorconfig` file added in #167 was not the latest version of the file.
> https://github.com/WordPress/twentynineteen/blob/master/.editorconfig

It should be updated to the latest from core:
> https://core.trac.wordpress.org/browser/trunk/.editorconfig

#Good-First-Issue

frankew:
Updates the editorconfig added in #167. Fixes #173

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -19,6 +19,3 @@ indent_size = 2
 
 [*.md]
 trim_trailing_whitespace = false
-
-[{*.txt,wp-config-sample.php}]
-end_of_line = crlf
--- a/.editorconfig
+++ b/.editorconfig
@@ -13,9 +13,12 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 indent_style = tab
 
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -13,9 +13,12 @@ insert_final_newline = true
...
 
-[{package.json,*.yml}]
+[*.yml]
 indent_style = space
 indent_size = 2
 
+[*.md]
+trim_trailing_whitespace = false
+
 [{*.txt,wp-config-sample.php}]
```
swissspidy: This theme repo will never have a `wp-config-sample.php` file in it. Plus the `*.txt` rule is very old and WordPress-specific for Windows people who open the WordPress readme file.

I suggest removing that part entirely.

kjellr:
Thanks, everyone!

PR #176, opened 2018-10-19
Remove the_title_attribute from function twentynineteen_post_thumbnail

carolinan:
Removes  `the_title_attribute` from the function twentynineteen_post_thumbnail, so that the authors own alt text is used instead.

WordPress.org username: poena


kjellr:
Thanks @carolinan!

Issue #178, opened 2018-10-19
Phpcs escaping issue


chetansatasiya:
there are many places where Phpcs is giving the escaping errors.

allancole:
Closing this as per this comment: https://github.com/WordPress/twentynineteen/pull/179#issuecomment-435957301

PR #179, opened 2018-10-19
added proper ecaping functions

chetansatasiya:
Hi Tem,

I am making this PR to resolve all the Phpcs errors.

Fix: #178 

Thanks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 .DS_Store
+.idea
+.idea/*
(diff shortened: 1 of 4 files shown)


ocean90:
Just like core, default themes don't escape translations when not necessary as those are trusted. See https://core.trac.wordpress.org/ticket/30724 for background.

allancole:
Closing this PR since we don’t escape translations in default themes. Thanks for contributing @chetansatasiya 👍

PR #183, opened 2018-10-19
Fix spacing in content-page.php

DannyCooper:


spencerfinnell:
A more consistent practice for default theme template files is to use the alternative syntax for control structures: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#brace-style

DannyCooper:
@spencerfinnell I've seen examples of both ways used within a single file - https://github.com/WordPress/twentyseventeen/blob/master/components/post/content.php

I opt for braces if there is no echo'ing of HTML.

kjellr:
The conditional here has been removed since this  PR was submitted, so I'm going to close this one for now. Feel free to re-submit if there are still improvements to be made here. Thanks!

PR #184, opened 2018-10-19
Update content-single.php

DannyCooper:
Reduce opening and closing PHP tags and fix spacing


spencerfinnell:
Were these done automatically with the current code standards?

DannyCooper:
@spencerfinnell This file in master is not currently formatted correctly, no.

PR #185, opened 2018-10-19
Fix alignment of classic editor iframe embeds

### LINKED ISSUE #175: iframes not centered in content
### Context

An article originally authored with the classic editor has this Instagram post, embedded using the regular link paste/embed functionality

https://www.instagram.com/p/BkAUuELlveW/?taken-by=eliorivero

### Expected

iframe to be centered

### Result

iframe was positioned towards left

captura de pantalla 2018-10-18 a la s 23 19 23

### Notes

While I see there is styling applied to iframes in general in [_media.scss](https://github.com/WordPress/twentynineteen/blob/be850d11ee31434ad700db518ff8570f82e65b9c/sass/media/_media.scss), this is not applied to this iframe because Instagram posts have some inline styles in the iframe.

nielslange:
Fixes #175 

<table>
<tr>
<td>Before:

175 - before
</td>
<td>After:

175 - after
</td>
</tr>
</table>

### DIFF
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -84,23 +84,6 @@
 	}
 }
 
-//! Overwrite inline styles of classic editor iframe embeds
-.entry-content > iframe,
-.entry-summary > iframe {
-
-	margin: 32px $size__spacing-unit !important;
-	max-width: calc(100vw - (2 * #{ $size__spacing-unit })) !important;
-
-	@include media(tablet) {
-		margin: 32px calc(2 * (100vw / 12)) !important;
-		max-width: calc(8 * (100vw / 12)) !important;
-	}
-
-	@include media(desktop) {
-		max-width: calc(6 * (100vw / 12)) !important;
-	}
-}
-
 .entry-content {
 
 	//! Audio
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -84,6 +84,23 @@
 	}
 }
 
+//! Overwrite inline styles of classic editor iframe embeds
+.entry-content > iframe,
+.entry-summary > iframe {
+
+	margin: 32px $size__spacing-unit !important;
+	max-width: calc(100vw - (2 * #{ $size__spacing-unit })) !important;
+
+	@include media(tablet) {
+		margin: 32px calc(2 * (100vw / 12)) !important;
+		max-width: calc(8 * (100vw / 12)) !important;
+	}
+
+	@include media(desktop) {
+		max-width: calc(6 * (100vw / 12)) !important;
+	}
+}
+
 .entry-content {
 
 	//! Audio
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -16,3 +16,4 @@ Version 1.0:
 @khleomix
 @audrasjb
 @nielslange
+@mmaumio
--- a/sass/site/primary/_posts-and-pages.scss
+++ b/sass/site/primary/_posts-and-pages.scss
@@ -202,5 +202,18 @@
 				max-width: calc(6 * (100vw / 12)) !important;
 			}
 		}
+
+		// Classic editor audio embeds.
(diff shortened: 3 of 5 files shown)


eliorivero:
Very nice, looks perfect now 👍Great work!

sass/blocks/_blocks.scss
```diff
@@ -84,6 +84,23 @@
 	}
 }
 
+//! Overwrite inline styles of classic editor iframe embeds
+.entry-content > iframe,
+.entry-summary > iframe {
+
+	margin: 32px $size__spacing-unit !important;
```
kjellr: In general this patch looks good, but is there any way we can accomplish this without the use of `!important`? There may not be in this case, but I always just make sure to ask. 🙂

joyously: You might just need to make the selector more specific.

nielslange: @kjellr and @joyously Believe me, I don't like `!important` statements either. Unfortunaterly, using the `!important` statement is the only way to fix this issue as I need to overwrite CSS that had been added directly to the iframe element:

> <iframe [...] style="background: white; max-width: 640px; width: calc(100% - 2px); border-radius: 3px; border: 1px solid rgb(219, 219, 219); box-shadow: none; display: block; margin: 0px 0px 12px; min-width: 326px; padding: 0px;"></iframe>

joyously: Seems like it would be a padding on `entry-content` instead of a margin on everything else.

kjellr: > @kjellr and @joyously Believe me, I don't like !important statements either. Unfortunaterly, using the !important statement is the only way to fix this particular issue, inline styles of the iframe element need to be overwritten

I did some digging as well, and came to the same conclusion. A last (but potentially weird?) resort would be to change the selector from `iframe` to `iframe[style]`, which would only target frames that have inline styles added to them. This would narrow the scope a little bit. Mind giving that a try?

> Seems like it would be a padding on entry-content instead of a margin on everything else.

The reason we don't currently apply margins to `entry-content` is so that we can more elegantly handle Gutenberg's wide & full alignment blocks. This is also used in the [Gutenberg Starter Theme](https://github.com/WordPress/gutenberg-starter-theme), and is a recommended method as per @jasmussen's [demo](https://codepen.io/joen/pen/zLWvrW) from the [Gutenberg docs](https://wordpress.org/gutenberg/handbook/extensibility/theme-support/#wide-alignments-and-floats). We could definitely work out a different solution (this isn't the way @allancole and I handled Gutenberg styles in [the Music theme](https://themeshaper.com/2018/05/22/music-a-gutenberg-powered-theme/) for instance), but it's been mostly working well except for this (relatively small) issue at this point.

nielslange: > Seems like it would be a padding on `entry-content` instead of a margin on everything else.

@joyously I’m not sure I understand what you meant. Would you mind explaining it to me? 

> I did some digging as well, and came to the same conclusion. A last (but potentially weird?) resort would be to change the selector from `iframe` to `iframe[style]`, which would only target frames that have inline styles added to them. This would narrow the scope a little bit. Mind giving that a try?

@kjellr Sure, that sounds like an elegant and save approach. 😉

joyously: > I’m not sure I understand what you meant. Would you mind explaining it to me?

In order for the content to have white space on both sides, margins are being applied to each kind of content. An alternative would be to apply padding to `entry-content` instead, and use negative margins for the things that are wide. 
But margins on all is the direction we are going.

kjellr:
I made a couple quick edits: 

- Added `[style]` to the `iframe` selector, as discussed.
- Moved the styles out of `_blocks.scss` and into `_posts-and-pages.scss`, since these are _technically_ not blocks. 
- Pulled in a fix for #198 as well, since the fix was very similar. 
- Added @cmmaumio to  `contributors.md`, since this involved his fix too. 🙂

Thanks everyone!

Fixes #175, Fixes #198

PR #186, opened 2018-10-19
Underline links within content

### LINKED ISSUE #15: undeline links in content
Links within content should be underlines, otherwise they are hard to see for people with colorblindness. 

bildschirmfoto 2018-10-16 um 16 27 05 kopie
bildschirmfoto 2018-10-16 um 16 27 05
bildschirmfoto 2018-10-16 um 16 28 13

nielslange:
Fixes #15 

<table>
<tr>
<td>Before:

15 - before
</td>
<td>After (dotted):

15 - after - dotted
</td>
<td>After (dashed):

15 - after - dashed
</td>
<td>After (solid):

15 - after - solid
</td>
<td>After (double):

15 - after - double
</td>
</tr>
</table>

### DIFF
--- a/sass/site/primary/_posts-and-pages.scss
+++ b/sass/site/primary/_posts-and-pages.scss
@@ -186,5 +186,18 @@
 				text-decoration: none;
 			}
 		}
+
+		a {
+			border-bottom: 1px dotted $color__link;
+		}
+
+		a:hover,
+		a:active {
+			border-color: $color__link-hover;
+		}
+
+		a:focus {
+			text-decoration: none;
+		}
 	}
 }
--- a/style.css
+++ b/style.css
@@ -68,8 +68,7 @@ html {
   line-height: 1.15;
   /* 1 */
   -webkit-text-size-adjust: 100%;
-  /* 2 */
-}
+  /* 2 */ }
 
 /* Sections
 	 ========================================================================== */
@@ -77,8 +76,7 @@ html {
  * Remove the margin in all browsers.
  */
 body {
-  margin: 0;
-}
+  margin: 0; }
 
 /**
  * Correct the font size and margin on `h1` elements within `section` and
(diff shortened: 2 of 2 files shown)


kjellr:
Thanks @nielslange. Due to the prevalent usage of solid lines throughout the theme, I learn towards the single solid line option (though I know that's boring 😄). 

Unless we add some ([fancy CSS to handle this](https://css-tricks.com/styling-underlines-web/)) I grudgingly suggest we use the standard `text-decoration: underline` instead of `border` though, so that it won't overlap/bump up against descenders: 

`border`
screen shot 2018-10-19 at 11 25 12 am

`text-decoration`
screen shot 2018-10-19 at 11 28 46 am

I'd love to hear thoughts from @melchoyce though. 🙂
 
---

Sidenote: I theoretically like the idea of having a 2px bottom border on links, so that it mimics the 2px borders we use throughout the theme. However, using our blue at 100% there seems a little too prominent visually:

screen shot 2018-10-19 at 11 13 32 am

... and lightening the color much more than that risks the underline not passing WCAG AA, so we can't do that. 

---

Also one minor note: we need to make sure that any treatment changes here are carried through to the editor styles as well. 👍

melchoyce:
I just went down a bit of a rabbit hole reading https://css-tricks.com/styling-underlines-web/. tl;dr we could probably do something more attractive than `text-decoration: underline;` but that is going to be by far the simplest solution.

kjellr:
> I just went down a bit of a rabbit hole reading https://css-tricks.com/styling-underlines-web/. tl;dr we could probably do something more attractive than text-decoration: underline; but that is going to be by far the simplest solution.

😄 That article is mesmerizing. Alright so let's do this: 

1. Update this PR to add a standard `text-decoration: underline;`. This will allow us to solve the a11y issue for now. 
2. Open a separate `enhancement` issue for more iteration on this style.  

Does that sound good, @nielslange?

nielslange:
@kjellr I just updated my PR and created #205 to have a separate discussion about future link styling.

nielslange:
Good catches, @kjellr!

Issue #143, opened 2018-10-18
Add support for submenu items


fabiankaegy:
I'm not sure how they would fit in this menu but I think it would be sad for a core theme not to include this functionality. At leaset one level deep. 
<img width="1920" alt="bildschirmfoto 2018-10-18 um 09 29 56" src="https://user-images.githubusercontent.com/20684594/47137968-624f9c00-d2b8-11e8-923d-0e6392a86853.png">

kjellr:
Yes! @allancole and I did not get a chance to implement this in the first draft we pushed up, but we should definitely support submenu items. 

Here's the mockup for large screens:

![menu-desktop](https://user-images.githubusercontent.com/1202812/47155364-8ad49780-d2b2-11e8-9c8e-9a20085fa617.png)

I'd love to morph that into something that feels spacious and mobile-native for small screens:

![menu-mobile-large](https://user-images.githubusercontent.com/1202812/47155399-a049c180-d2b2-11e8-97f5-f1d1583fd26c.png)

Here's a GIF of that in action: 

![menu-mobile](https://user-images.githubusercontent.com/1202812/47155425-afc90a80-d2b2-11e8-8b51-521ec065f9bf.gif)

We also had the concept of a fixed mobile menu on scroll — not sure if we'll pursue that (I think the direction in that GIF feels more modern), but here it is in case it sparks more discussion:

<img width="566" alt="screen shot 2018-10-18 at 8 44 19 am" src="https://user-images.githubusercontent.com/1202812/47155476-cff8c980-d2b2-11e8-91a4-0de12d39de34.png">

I'm going to be doing some thinking about the menu in general, so all this may change just a little bit... But in general: Yes, we need submenu support. 👍

fabiankaegy:
I really like the design shown in the gif. 👍

melchoyce:
One quick note on mobile — there's a bit of cognitive dissonance that happens when you click a down arrow, and the panel comes in from the side. It's a little disorienting at first. Maybe change the direction of the arrow, or the direction the panel comes in from?

kjellr:
> One quick note on mobile — there's a bit of cognitive dissonance that happens when you click a down arrow, and the panel comes in from the side. It's a little disorienting at first. Maybe change the direction of the arrow, or the direction the panel comes in from?

That's a great point. Changing the direction seems like a simple solution, though we could even swap out the icon entirely if we wanted to.

kjellr:
[...]

PR #192, opened 2018-10-19
Add Submenu Support

allancole:
Addresses #143 to make the theme support sub-menus. 

Some notes: 

• Supports tabbed browsing and `:focus` states for accesibility
• Submenus are displayed correctly on mobile (including iOS), and desktop
• Still needs some styling to make the icon look nicer
• It doesn’t yet match the design that Kjell introduced in #143 but gets us a bit closer to it


allancole:
This is a lo-fi version of the menu. Merging it in now so that submenus are covered, but it will likely be improved in another PR to better support `i11n` and expected menu appearance/behavior for `a11y`.

PR #193, opened 2018-10-19
Escaping author display name in get_the_author() in author-info.php

mmaumio:
esc_url() function not being used in author-info.php to escape the output of author display name with get_author_name()

### DIFF
--- a/template-parts/post/author-info.php
+++ b/template-parts/post/author-info.php
@@ -10,7 +10,7 @@ if ( (bool) get_the_author_meta( 'description' ) ) : ?>
 <div class="author-description">
 	<h2 class="author-title">
 		<span class="author-heading"><?php _e( 'Published by', 'twentynineteen' ); ?></span>
-		<?php echo get_the_author(); ?>
+		<?php echo esc_html( get_the_author() ); ?>
 	</h2>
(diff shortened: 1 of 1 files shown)


template-parts/post/author-info.php
```diff
@@ -10,7 +10,7 @@
 <div class="author-description">
 	<h2 class="author-title">
 		<span class="author-heading"><?php _e( 'Published by', 'twentynineteen' ); ?></span>
```
spencerfinnell: This is not properly translatable. It should be:

```suggestion
		<span class="author-heading"><?php echo esc_html( sprintf( __( 'Published by %s', 'twentynineteen' ), get_the_author() ) ); ?></span>
```

spencerfinnell: With a translator note.

mmaumio: Thanks for your input. I'll check and apply the suggestion.

allancole: Yup, make that change and we can go ahead and merge this ;-)

mmaumio:
Implemented @spencerfinnell suggested changes along with mine.
I hope this will work.

Thank You

PR #197, opened 2018-10-19
Fixed audio player overlaping post/page container

mmaumio:
Using self hosted Audio file with audio shortcode using Media Element js on any post/page overlaps the entire post/page content area.
Fixed the issue here.

WordPress.org: mmaumio

### DIFF
--- a/classes/class-twentynineteen-walker-comment.php
+++ b/classes/class-twentynineteen-walker-comment.php
@@ -69,12 +69,12 @@ class TwentyNineteen_Walker_Comment extends Walker_Comment {
 						</a>
 						<?php
 							$edit_comment_icon = twentynineteen_get_icon_svg( 'edit', 16 );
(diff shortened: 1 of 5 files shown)


nielslange:
@mmaumio The fix looks great. Could you also add the change to the SCSS file? Otherwise, your change would get lost the first time the SCSS file gets rendered.

mmaumio:
> @mmaumio The fix looks great. Could you also add the change to the SCSS file? Otherwise, your change would get lost the first time the SCSS file gets rendered.

Thank You.
Just did.

Please let me know if there's anything else.

kjellr:
👋 I got fix in alongside #185, since it turned out they actually needed mostly the same code. 🙂 

Thanks for your help! I've [added you](
https://github.com/WordPress/twentynineteen/pull/185/commits/13371b31ee8e3bc4a9391996a32cffa5ee8eb322) to the contributors file!

mmaumio:
Thank you for accepting my code and adding me to the Contributors.
It's a privilege :)

PR #199, opened 2018-10-19
Add minimum-scale=1 to meta viewport

westonruter:
The use of `minimum-scale=1` is [required](https://www.ampproject.org/docs/fundamentals/spec#vprt) for valid AMP documents:

> contain a `<meta name="viewport" content="width=device-width,minimum-scale=1">` tag inside their head tag. It's also recommended to include `initial-scale=1`.

The theme is already doing the required`width=device-width` and the recommended `initial-scale=1`, but it is not doing `minimum-scale=1`. Any reason not to include it?

### DIFF
--- a/header.php
+++ b/header.php
@@ -14,7 +14,7 @@
 <html <?php language_attributes(); ?>>
 <head>
 	<meta charset="<?php bloginfo( 'charset' ); ?>" />
-	<meta name="viewport" content="width=device-width, initial-scale=1" />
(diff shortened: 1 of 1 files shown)


peterwilsoncc:
> Any reason not to include it?

I'm not a fan of limiting scaling in either direction. 

In the event an embed or other replaced element introduces horizontal scaling it's helpful to zoom below 1.0 to view the entire thing.

westonruter:
This shouldn't be required by AMP in the future anyway. See https://github.com/ampproject/amphtml/issues/18896

Issue #104, opened 2018-10-17
Styling sticky posts


LittleBigThing:
Right now, a sticky post is visually not distinguishable from other posts.

Consider adding some specific styling or some kind of label ("featured" as in other default themes, a star (mind accessibility)?, ...) to sticky posts.

khleomix:
@LittleBigThing do you have anything specific in mind with regards to the design of the sticky posts? Twenty Seventeen has a thumb tack icon, Twenty Sixteen and Twenty Fifteen both have the `<span class="sticky-post">Featured</span>`

Are we going to do something similar? @kjellr do we have a design for this? Will take this up.

LittleBigThing:
Nothing specific, @khleomix.

- It could be as little as adding the label "Featured" to the entry meta data under the post, maybe with a star as an icon.
- Or adding a larger (star) icon left to the title in the margin (with sufficient attention to a11y), much like to custom logo can be added next to the site title.
- Or &mdash; and I might prefer this option &mdash; style the **first** sticky post in the list of posts like a single post is styled (also including a 'Featured' label). So, with the wide, full-screen featured image and the post title above it.

I would love to join in helping to realise one of these options or another cool idea.

khleomix:
@LittleBigThing I'll do some mockups tomorrow and we can start from there.

khleomix:
@LittleBigThing here are some mockups, I do prefer having the first sticky post styled like a single post. We'll probably need to set a default background for those that don't have featured images though...

![starred](https://user-images.githubusercontent.com/1948204/47233353-05153100-d405-11e8-94f9-1e231e20a437.jpg)
![post-meta-featured](https://user-images.githubusercontent.com/1948204/47233354-09414e80-d405-11e8-8a0c-b56289a70e1e.jpg)
![post-header-featured](https://user-images.githubusercontent.com/1948204/47233356-0ba3a880-d405-11e8-91e1-ba8968cd85f4.jpg)
![most-recent-featured](https://user-images.githubusercontent.com/1948204/47233360-0d6d6c00-d405-11e8-8143-ee40df9a921e.jpg)

LittleBigThing:
Nice, thanks! I like them all and I prefer also the one that is styled as a single post. Although I am wondering if it isn’t confusing havind the other posts below it. Also, we should determine what we would do if there are more sticky posts.

Anyway, we need probably some design feedback, @kjellr?

khleomix:
[...]

PR #201, opened 2018-10-19
add label featured to sticky post

LittleBigThing:
Adding the label 'featured' above the title of sticky post(s) on the post list page as discussed in #104

### DIFF
--- a/sass/site/primary/_posts-and-pages.scss
+++ b/sass/site/primary/_posts-and-pages.scss
@@ -2,6 +2,18 @@
 	display: block;
 }
 
+.sticky-post {
+	background-color: #0073aa;
+	color: #fff;
+	display: inline-block;
+	line-height: 1;
+	padding: .25rem;
+	position: absolute;
+	text-transform: uppercase;
+	top: -1rem;
+	z-index: 1;
+}
+
 .updated:not(.published) {
 	display: none;
 }
--- a/sass/site/primary/_posts-and-pages.scss
+++ b/sass/site/primary/_posts-and-pages.scss
@@ -3,7 +3,7 @@
 }
 
 .sticky-post {
-	background-color: #0073aa;
+	background: #0073aa;
 	color: #fff;
 	display: inline-block;
 	line-height: 1;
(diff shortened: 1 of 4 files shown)


sass/site/primary/_posts-and-pages.scss
```diff
@@ -2,6 +2,18 @@
 	display: block;
 }
 
+.sticky-post {
```
kjellr: This looks good, thanks @LittleBigThing! My only suggestion would be to change this `sticky-post` class name to something that's slightly more descriptive: `sticky-label` or `sticky-post-label` for instance. (At first read, I wrongly assumed `sticky-post` was a wrapper for the entire post).

joyously: Did you know that titles can contain markup? So if it has some word in bold, it won't be any different.

kjellr: > Did you know that titles can contain markup? So if it has some word in bold, it won't be any different.

To make  sure I'm clear: you're referring to the post title? Yes, that can contain limited markup. However, I don't believe that is related to this specific PR. The only element we're making bold here is the "FEATURED" label above the post title. Since we know exactly what that says, I don't think it needs any special considerations. 

As an aside though, we actually do have styles in place to make sure that bold post titles are displayed in a bolder weight:

screen shot 2018-10-19 at 10 44 28 pm

joyously: Cool. I couldn't tell from your comment which part should be bold.

LittleBigThing:
Thanks for the feedback, @kjellr!

I have made the label bold. It looks nice, a bit more striking and it matches the title style, indeed.

About the class name. I am happy to modify `.sticky-post` into `.sticky-post-label`, but I wanted to mention that `.sticky-post` was used in Twenty Fifteen as well as in Twenty Sixteen. Just so we can make a final decision. :-)

screenshot 2018-10-20 at 10 25 42

kjellr:
> About the class name. I am happy to modify .sticky-post into .sticky-post-label, but I wanted to mention that .sticky-post was used in Twenty Fifteen as well as in Twenty Sixteen. Just so we can make a final decision. :-)

You're right! Good call. Let's stick with that. 🙂 This is all set to merge in now then. Thanks again!

PR #202, opened 2018-10-19
❌ REMOVE: Jetpack Support

asharirfan:
This PR is for removing all the references of Jetpack from Twenty Nineteen default theme.

CC @allancole


nielslange:
@asharirfan May I ask why you removed the support for Jetpack? Having in mind that Jetpack has 5+ million active installations and that some Jetpack features, see #210, will look bad using Twenty Nine I don’t think we should remove Jetpack support.

asharirfan:
@nielslange See https://github.com/WordPress/twentynineteen/pull/55#issuecomment-431433836

nielslange:
I see. However, that’ll mess up both the Akismet messages as well as the Jetpack’s related posts. 

@westonruter & @allancole Based on your discussion in #55 I wonder what your thoughts concerning #210 are.

westonruter:
Historically the core themes have not had direct support for any plugins.

nielslange:
> Historically the core themes have not had direct support for any plugins.

Thank for your explanation, @westonruter. However, I still wonder how this issue should be treated then. Clearly, the Twenty Nineteen themes break the way Jetpack's related posts look like. In y opinion that's also nothing that should be fixed on Jetpack's side. What advice would you give your customers? Adding some custom CSS via customizer to fix this issue?

westonruter:
Jetpack itself ships with compat styles for core themes: https://github.com/Automattic/jetpack/tree/master/modules/theme-tools/compat

nielslange:
Thanks, @westonruter! 😀

PR #203, opened 2018-10-19
Only show site-title as a h1 when it's the homepage/frontpage

### LINKED ISSUE #5: H1 tag in site-branding
Hello,

Like many other themes, I think we shouldn't use H1 tag in site-branding.php if we are not on the website homepage (for both front page or blog post page).

kjellr:
For better SEO, and for preventing the appearance of more than one h1 on the page. Fixes #5, and addresses one checkbox  in #41.

This is basically just an updated version of #3 (it was quicker than solving the merge conflicts in that branch), so @audrasjb gets all the props for this one. 🙌

### DIFF
--- a/sass/typography/_headings.scss
+++ b/sass/typography/_headings.scss
@@ -8,6 +8,7 @@ h2.author-title,
 .comments-title,
 .comment-author .fn,
 .no-comments,
+.site-title,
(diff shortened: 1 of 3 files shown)


allancole:
Yup, looks good to me @kjellr 👍

PR #212, opened 2018-10-20
Fix comments pagination links

### LINKED ISSUE #211: placement of paged comments link
The placement of the paged comments link is a bit off. It needs the same margin as everything else, and it would make better sense if it were above the comment form.

paged-comments

nielslange:
Fixes #211 

<table>
<tr>
<td>Before:

210 - before</td>
<td>After:

210 - after</td>
</tr>
</table>


joyously:
Looks good, but it would still be more useful to have the links near the comments instead of after the comment form.

PR #214, opened 2018-10-20
Fix overlapping Gravatar when leaving a comment

### LINKED ISSUE #213: Placement of comment user avatar/gravatar is off
This only happens on posts/pages that don't have comments yet. The comment user avatar's placement is off and covers the top part of the comment form. Happens on both logged out and logged in users.

comment-logged-out
comment-logged-in

nielslange:
Fixes #213

<table>
<tr>
<td>Before:

213 - before</td>
<td>After:

213 - after</td>
</tr>
</table>

### DIFF
--- a/inc/template-functions.php
+++ b/inc/template-functions.php
@@ -114,6 +114,15 @@ function twentynineteen_can_show_post_thumbnail() {
 }
 
 /**
+ * Determines the estimated time to read a post, in minutes.
+ */
+function twentynineteen_get_estimated_reading_time() {
+	$content = get_post_field( 'post_content', get_the_ID() );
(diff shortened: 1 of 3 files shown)


inc/template-functions.php
```diff
@@ -113,6 +113,15 @@ function twentynineteen_can_show_post_thumbnail() {
 	return ! post_password_required() && ! is_attachment() && has_post_thumbnail();
 }
 
+/**
```
kjellr: @nielslange This doesn't seem like it should be included in this PR. Can you remove this function to sync back up with master?

nielslange: @kjellr Good catch! This code seemed to slip in there accidentally. I just removed it and updated this PR.

PR #221, opened 2018-10-20
Add instruction for compiling SCSS

jaymanpandya:
I had a doubt of how to compile SASS, so I went to Slack to see if I can find some information about it and I found a conversation (https://wordpress.slack.com/archives/C02RP4VMP/p1539871080000100) about it. 

Hence,  added the information shared on Slack in the contributing document so that people can quickly get started.


allancole:
Thanks for sorting this out but we’ll have to close it since it’s addressed in #169 .

PR #233, opened 2018-10-22
remove background featured image

LittleBigThing:
Remove the background image that is added to the featured image's anchor element on post lists


LittleBigThing:
#234

PR #235, opened 2018-10-22
Update “Requires at least” and add it to stylesheet

### LINKED ISSUE #207: readme and version required
Current usage is that themes do not have a "Requires at least" tag.
There are core tickets open to add it, and also tickets to add a standard readme file to themes (instead of just using style.css headers).
This theme's readme.md has "Requires at least  4.9.8", but it actually only requires 4.9.6 because of  `the_privacy_policy_link()`. Beyond that, it only needs 4.5 for the custom logo.
Is something else going to be added from 4.9.8?

nielslange:
Fixes #207

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 **Contributors:** the WordPress team  
 **Requires at least:** WordPress 4.9.8  
-**Tested up to:** WordPress 4.9.8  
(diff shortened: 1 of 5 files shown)


allancole:
Thanks for fixing this @nielslange ;-)

PR #244, opened 2018-10-23
Replace the_content with the_excerpt when necessary in content.php

audrasjb:
Fixes: #89 and closes PR #96.


DannyCooper:
@audrasjb Wouldn't you want to keep the code inside the_content()? It generates a screen-reader-friendly, translatable 'read more'.

audrasjb:
@DannyCooper I think we don't need that since the_content() is used after is_singular() condition, isn't it?

I can't figure what is the goal: are you thinking of Read More Block. I think it generates an accessible markup by itself :)

DannyCooper:
I'm not sure it does:

```
	if ( null === $more_link_text ) {
		$more_link_text = sprintf(
			'<span aria-label="%1$s">%2$s</span>',
			sprintf(
				/* translators: %s: Name of current post */
				__( 'Continue reading %s' ),
				the_title_attribute( array( 'echo' => false ) )
			),
			__( '(more&hellip;)' )
		);
	}
```

It doesn't have `<span class="screen-reader-text"> "%s"</span>` by default

joyously:
That is totally different code, isn't it?
`the_content()` only outputs the Read More if it a list page and there is a `<!--nextpage-->` tag.
I wondered if full content or excerpts could be a user option, but that is a different issue, maybe.

DannyCooper:
@joyously This template part is used on a list page (index.php) and it's when `<!--more-->` is used the code (which I'm saying shouldn't be removed) will be shown.

joyously:
I'm sorry --- I typed `<!--nextpage-->` when I meant `<!--more-->`.
Where exactly is the code you are talking about?
If the_content() is in an if statement for singular, then the More is not used.
Is there somewhere else it is called?

DannyCooper:
Ah yes, my mistake. I didn’t realise this will only be used on singular pages.

joyously:
What about a Customizer option?
_And for the record, I don't think it removes the editor from the "index feed". The editor was never in the feed. The content is, though, and you have to scroll a lot._

allancole:
> What about a Customizer option?

We probably shouldn’t add a theme option to the customizer for something that is already achievable with built-in WP functionality. The `<!-- more -->` tag exists so that users can shorten their posts when they want to on a per-post basis. And especially with the Gutenberg editor, that more tag is more easy to discover than in a theme option. 

> And for the record, I don't think it removes the editor from the "index feed". The editor was never in the feed. The content is, though, and you have to scroll a lot.

You are correct. To clarify, I mean the _output_ of the Gutenberg editor—not the editor itself.

joyously:
> We probably shouldn’t add a theme option to the customizer for something that is already achievable with built-in WP functionality.

Options are great for when using built-in functionality means you have to change every post, though. Lots of legacy content around.
Themes should be flexible about the part that they do. This has nothing to do with the editor.

kjellr:
I'm going to close this one, since #89 has been closed via #361.

PR #249, opened 2018-10-23
Remove unused IE stylesheet enqueue.

### LINKED ISSUE #248: Unused ie-theme stylesheet
On every page load, I noticed unnecessary calls to an unexisting CSS `/css/ie.css` file resulting in 404.

https://github.com/WordPress/twentynineteen/blob/bf8eabe5c8965237deda50635abbeb80ece96df4/functions.php#L138

Is there a specific reason for that?

kjellr:
Removes an unused IE stylesheet. I don't think this is intentionally included in here, but CC-ing @allancole to confirm. 

Fixes #248


allancole:
Yup, that can be removed.

PR #240, opened 2018-10-22
remove unused variable with bg image url

LittleBigThing:
I was a variable forgotten with the post thumbnail url for the background image in PR #233 

Related to #234

PR #250, opened 2018-10-23
Generate post thumbnail variable for single post pages.

kjellr:
#240 removed the generation of `$post_thumbnail` variable in post lists. But it also seems to have removed the overlay from single post pages as well.

I'm not 100% sure why that happened, but generating the `$post_thumbnail` for `is_singular()` posts seems to fix. 

cc @allancole for a review.

### DIFF
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -140,6 +140,7 @@ if ( ! function_exists( 'twentynineteen_post_thumbnail' ) ) :
 		}
 
 		if ( is_singular() ) :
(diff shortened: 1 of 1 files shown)


kjellr:
Interesting! I guess we're all set. Just re-checked and it's working now. 🔮

LittleBigThing:
Strange, sorry about that...

kjellr:
No problem! I'm not 100% sure that was related to your PR anyway. 🙂

PR #251, opened 2018-10-23
Replace “Tested up to” with “Requires at least”

nielslange:
@allancole I messed up PR #235 for issue #207! 🙄 

Instead of `Requires at least` I changed the version number for `Tested up to` and added that one to the stylesheets. I just corrected that issue and since I run `npm run build` that part had also been added to the file `style-rtl.css`. Sorry for the inconvenience though. 🤭


allancole:
Good catch! Thanks 👍

PR #253, opened 2018-10-23
Add spacing to page links

### LINKED ISSUE #209: page links run together
The page links are a little close together, so it looks like one number. Perhaps use more space or commas between them?

page-links

nielslange:
Fixes #209 

<table>
<tr>
<td>Before:

209 - before</td>
<td>After (underlined):

209 - after underlined</td>
<td>After (not underlined):

209 - after not underlined</td>
</tr>
</table>

@joyously and @kjellr Due to #186 the page links would be underlined. I created an underlined version and a non-underlined version (as seen in the screenshot). I wonder if we should underline both these page links as well as the other pagination links due to accessibility reasons. 🤔


kjellr:
Thanks, @nielslange — my personal inclination is to say that these _don't_ need an underline because they're not content, and because the normal pagination links also do not have underlines. But I'll hold off for more thoughts before making that call.

nielslange:
> Thanks, @nielslange — my personal inclination is to say that these _don't_ need an underline because they're not content, and because the normal pagination links also do not have underlines. But I'll hold off for more thoughts before making that call.

I run a quick colourblindness test for the implementation without underlined links and the result looks good. Based on the `font-weight` a colourblind person should be able to distinguish between linked and unlinked page numbers:

209 - after not underlined greyscale

sass/typography/_headings.scss
```diff
@@ -117,6 +117,11 @@ h6 {
 	font-weight: bold;
 }
 
+.entry-content .page-links .post-page-numbers {
```
kjellr: @nielslange Would you mind giving this another look? I don't see the `post-page-numbers` class in my tests, so things are looking like this: 

screen shot 2018-10-29 at 2 10 37 pm

allancole: @nielslange This fix should appear in the `sass/site/primary/_posts-and-pages.scss` sass file. The `_headings.scss` is specifically for grouping typographically similar elements and should only address font-related styles. Fix that and the point @kjell mentioned and we can get this merged in. 👍

allancole:
Update the styles for this

image

PR #255, opened 2018-10-23
Improve archive header

DannyCooper:
https://wordpress.slack.com/archives/C02RP4VMP/p1540306253000100


sass/site/primary/_archives.scss
```diff
@@ -33,6 +33,12 @@
 	}
 }
 
+.archive .page-header .page-description {
+	display: block;
+	color: #111;
```
kjellr: This PR looks good, thanks @DannyCooper. Let's just swap this color value out with our Sass variable `$color__text-main`, and then we'll be ready to go. 👍

DannyCooper:
Thanks for the review @kjellr, I've now added the variable.

PR #262, opened 2018-10-24
Consolidate editor styles

kjellr:
❗️ This shouldn't be merged until https://github.com/WordPress/gutenberg/pull/10956 is officially part of Gutenberg (Planned for the Gutenberg 4.2 Release)❗️ 

---

A followup to #22.

This PR removes our secondary `style-editor-frame.css` file, and consolidates all of our editor styles into `style-editor.css`. Initially, the standard editor stylesheet was limited in scope to individual blocks, which meant we were unable to style the general alignment of the page, as well as items like the page title and default appender.

The scope of `style-editor.css` is being expanded in WordPress/gutenberg#10956 , so we're now able to style those elements in a single stylesheet.

Some minor style tweaks were necessary to make our original styles work in this context, but for the most part this is identical to #22.


kjellr:
@allancole This should be good to go. 

To test: 

- Install [Gutenberg 4.2-RC1](https://make.wordpress.org/core/2018/10/31/whats-new-in-gutenberg-31st-october-2/)
- Make sure the editor alignment styles are still in place, as well as the editor post title + default appender font styles.
 - Make sure changing the color still works via the customizer (since I renamed that stylesheet).

PR #263, opened 2018-10-24
Adjust main menu

### LINKED ISSUE #32: Make the "links" look more like a menu
I've had a look at the mockups and the "menu" doesn't actually look like a menu, which can be a real problem for some users.

nielslange:
Fixes #32 partially

What I've done:

- Display main menu in full width
- Removed commas between menu items
- Removed dot behind last menu item
- Show drop-down menu below parent menu item

What needs to be done:

- Implement priority+ navigation pattern as described in https://github.com/WordPress/twentynineteen/issues/32#issuecomment-431892606

<table>
<tr>
<td>Before:

32 - before</td>
<td>After:

32 - after</td>
</tr>
</table>

### DIFF
--- a/sass/navigation/_menus.scss
+++ b/sass/navigation/_menus.scss
@@ -2,11 +2,7 @@
 
 .main-navigation {
 
-	display: inline;
-
-	@include media(tablet) {
-		display: block;
-	}
+	display: block;
 
 	body.page & {
 		display: block;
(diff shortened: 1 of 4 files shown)


kjellr:
Thanks @nielslange! This looks good. I pushed one very minor update (adding a little top margin to match the top margin of the main menu to the bottom margin), but this is good to go. 🎉

PR #274, opened 2018-10-24
Add missing text domain and escaping to comment author text in class-twentynineteen-walker-comment.php

joshfeck:
The changes in this branch move out html tags that are within some of the i18n-ready strings. These changes follow the suggested practice of "Do not put unnecessary HTML markup into the translated string" from 
https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#best-practices-for-writing-strings

My wp.org username is @builtbynorthby.

Thanks for reviewing.


WPprodigy:
Came across the same here: https://github.com/WordPress/twentynineteen/blob/7eb93d5f79eedb168f72ccfeddbfcf6f0f129875/classes/class-twentynineteen-walker-comment.php#L57-L60

Would be nice to move out the HTML from there for simplicity for translators, and also allow simpler escaping of the translated text.

joshfeck:
Nice catch @WPprodigy

joshfeck:
@grappler Thanks for the review.

classes/class-twentynineteen-walker-comment.php
```diff
@@ -52,10 +52,12 @@ protected function html5_comment( $comment, $depth, $args ) {
 								printf( '<span class="post-author-badge" aria-hidden="true">%s</span>', twentynineteen_get_icon_svg( 'check', 24 ) );
 							}
 
-							/* translators: %s: comment author link */
+							/* translators: %1$s: comment author link */
```
grappler: You would need to update the comments too.

classes/class-twentynineteen-walker-comment.php
```diff
@@ -52,10 +52,12 @@ protected function html5_comment( $comment, $depth, $args ) {
...
 							printf(
-								__( '%s <span class="screen-reader-text says">says:</span>' ),
-								sprintf( '<b class="fn">%s</b>', get_comment_author_link( $comment ) )
+								esc_html__( '%1$s %2$ssays:%3$s' ),
```
grappler: I am not sure if string is easier to read and translate with the many variables.

joshfeck: It's definitely less easy to read compared to the prior commit, but it does allow for changing the order to grammatically correct to `Says Sam:`.

With regards to the usage of esc_html__() breaking the html, are you sure? The reason I ask is because the esc_html__() function executes first. So it is only escaping this:`%1$s %2$ssays:%3$s` (or whatever gets passed in from gettext). Then after that, printf() takes the string and swaps in the html for the placeholders.

tn3rb: @grappler 

> re:  The escaping is going to cause the HTML to break.

no that won't happen because the call to `esc_html__()` happens **within** the call to `sprintf()`
so `esc_html__()` executes first and its result is passed to `sprintf()`.

the `esc_html__()` is **only** there to protect against HTML injected by the translators,
it will not escape the HTML added by the `sprintf()`

dimadin:
As a translator, I'm opposed to this. To quote this best practice:
> Do not put **unnecessary** HTML markup into the translated string

Sometimes you can't avoid having HTML. There [are](https://translate.wordpress.org/projects/wp/dev/admin/network/sr/default?filters%5Bterm%5D=%3C&filters%5Buser_login%5D=&filters%5Bstatus%5D=current_or_waiting_or_fuzzy_or_untranslated&filter=Filter&sort%5Bby%5D=priority&sort%5Bhow%5D=desc) [many](https://translate.wordpress.org/projects/wp/dev/sr/default?filters%5Bterm%5D=%3C&filters%5Buser_login%5D=&filters%5Bstatus%5D=current_or_waiting_or_fuzzy_or_untranslated&filter=Filter&sort%5Bby%5D=priority&sort%5Bhow%5D=desc) [examples](https://translate.wordpress.org/projects/wp/dev/admin/network/sr/default?filters%5Bterm%5D=%3C&filters%5Buser_login%5D=&filters%5Bstatus%5D=current_or_waiting_or_fuzzy_or_untranslated&filter=Filter&sort%5Bby%5D=priority&sort%5Bhow%5D=desc) in WordPress core.

Examples like this are harder to translate and read, as @grappler already said, I don't remember instance when they were ever like this in WordPress core or default themes:
https://github.com/WordPress/twentynineteen/blob/8b2cab44112c753cb537ea4cd380eae79f1a19c0/classes/class-twentynineteen-walker-comment.php#L57-L60

For strings with HTML element with `screen-reader-text` class, I think that translators got used to them. Some of examples are already [in](https://translate.wordpress.org/projects/wp-themes/twentyseventeen/sr/default?filters%5Bterm%5D=screen&filters%5Buser_login%5D=&filters%5Bstatus%5D=current_or_waiting_or_fuzzy_or_untranslated&filter=Filter&sort%5Bby%5D=priority&sort%5Bhow%5D=desc) Twenty Seventeen.

See #217 for how I worked to reduce HTML in strings, particularly examples in [`inc/template-tags.php` file](https://github.com/WordPress/twentynineteen/pull/217/files#diff-5915276465e785d5f7897d71685fca5fL41). In my opinion, this is perfect example of above mentioned best practice.

classes/class-twentynineteen-walker-comment.php
grappler: > The escaping is going to cause the HTML to break

Sorry I miss read the code. It should not have an effect. 

> It's definitely less easy to read compared to the prior commit,

I meant it was less easier to read then with the HTML tags.

ocean90:
I agree with the provided feedback by @dimadin. None of the changes are necessary and make it harder for translators to provide proper translations. I suggest to close this as wontfix.

mnelson4:
I realize I probably won't change the decision. But just adding a vote in favour of @joshfeck's PR here and avoiding HTML in translated strings, despite there being lots of cases where they exist in WP core (and we can't change them without breaking translations).

I appreciate @dimadin providing a ton of examples and previous work. But yes, IMO a string like 

```
sprintf(
			/* translators: %s: post author */
			__( '<span class="screen-reader-text">Posted by</span>%s', 'twentyseventeen' ),
			'<span class="author vcard"><a class="url fn n" href="' . esc_url( get_author_posts_url( get_the_author_meta( 'ID' ) ) ) . '">' . esc_html( get_the_author() ) . '</a></span>'
		);
```

would be better translated using placeholders for HTML tags, eg

```
sprintf(
			/* translators: %1$s: opening tag, %2$s: closing tag, %3$s: post author */
			__( '%1$sPosted by%2$s%3$s', 'twentyseventeen' ),
			'<span class="screen-reader-text">',
			'</span>',
			'<span class="author vcard"><a class="url fn n" href="' . esc_url( get_author_posts_url( get_the_author_meta( 'ID' ) ) ) . '">' . esc_html( get_the_author() ) . '</a></span>'
		);
```

Plus, this way translators won't be tempted to translate the contents of the HTML tag, and are less likely to mess up the HTML.

But again, I realize my vote probably won't change this decision. I'm just adding an extra data point.

tn3rb:
> > Do not put unnecessary HTML markup into the translated string
> 
> Sometimes you can't avoid having HTML. There are many examples in WordPress core.

Actually you CAN avoid having unnecessary HTML markup in the translated string... by properly using placeholders! Exactly as is being done in this PR!

Maybe experienced translators have gotten used to dealing with markup but what about someone attempting a translation for the first time?

I too am not translator, but I can't imagine why you would prefer having to deal with HTML over a few placeholders that have been properly commented!

crunnells:
I'm gonna agree with @allancole and say that I'm in favor of this change. The code legibility and cleaner translation strings are a clear improvement here.

header.php
```diff
@@ -50,16 +50,9 @@
...
-											array(
-												'span' => array(
-													'class' => array(),
-												),
-											)
-										),
-										get_the_title()
+										/* translators: %s: Name of current post. Only visible to screen readers. */
+										esc_html__( 'Edit %s', 'twentynineteen' ),
+										'<span class="screen-reader-text">' . get_the_title() . '</span>'
```
ocean90: This change needs to be reverted to support proper translations in all languages. See https://core.trac.wordpress.org/ticket/31979 and  https://github.com/WordPress/twentysixteen/pull/383 for background.

inc/template-tags.php
```diff
@@ -110,16 +110,9 @@ function twentynineteen_entry_footer() {
...
-					array(
-						'span' => array(
-							'class' => array(),
-						),
-					)
-				),
-				get_the_title()
+				/* translators: %s: Name of current post. Only visible to screen readers. */
+				esc_html__( 'Edit %s', 'twentynineteen' ),
+				'<span class="screen-reader-text">' . get_the_title() . '</span>'
```
ocean90: This change needs to be reverted to support proper translations in all languages. See https://core.trac.wordpress.org/ticket/31979 and  https://github.com/WordPress/twentysixteen/pull/383 for background.

template-parts/content/content-none.php
```diff
@@ -20,16 +20,12 @@
...
-						),
-					)
-				) . '</p>',
-				esc_url( admin_url( 'post-new.php' ) )
+				'<p>' .
+					/* translators: %s: link to WP admin new post page. */
+					esc_html__( 'Ready to publish your first post? %1$sGet started here%2$s.', 'twentynineteen' )
+				. '</p>',
+				'<a href="' . esc_url( admin_url( 'post-new.php' ) ) . '">',
+				'</a>'
```
ocean90: This change doesn't really make the code more readable. Also, the translators comment is not correct since there is no `%s` placeholder.

Worth to mention that `Ready to publish your first post? <a href="%1$s">Get started here</a>.` is a string that's already used by many themes.

template-parts/content/content-page.php
```diff
@@ -36,16 +36,9 @@
...
-						array(
-							'span' => array(
-								'class' => array(),
-							),
-						)
-					),
-					get_the_title()
+					/* translators: %s: Name of current post. Only visible to screen readers */
+					esc_html__( 'Edit %s', 'twentynineteen' ),
+					'<span class="screen-reader-text">' . get_the_title() . '</span>'
```
ocean90: This change needs to be reverted to support proper translations in all languages. See https://core.trac.wordpress.org/ticket/31979 and WordPress/twentysixteen#383 for background.

template-parts/content/content-single.php
```diff
@@ -32,16 +32,9 @@
...
-							array(
-								'span' => array(
-									'class' => array(),
-								),
-							)
-						),
-						get_the_title()
+						/* translators: %s: Name of current post. Only visible to screen readers. */
+						esc_html__( 'Edit %s', 'twentynineteen' ),
+						 '<span class="screen-reader-text"> ' . get_the_title() . '</span>'
```
ocean90: This change needs to be reverted to support proper translations in all languages. See https://core.trac.wordpress.org/ticket/31979 and WordPress/twentysixteen#383 for background.

template-parts/content/content-single.php
```diff
@@ -56,16 +49,9 @@
...
-					array(
-						'span' => array(
-							'class' => array(),
-						),
-					)
-				),
-				get_the_title()
+				/* translators: %s: Name of current post. Only visible to screen readers */
+				esc_html__( 'Continue reading %s', 'twentynineteen' ),
+				'<span class="screen-reader-text">"' . get_the_title() . '"</span>'
```
ocean90: This change needs to be reverted to support proper translations in all languages. See https://core.trac.wordpress.org/ticket/31979 and WordPress/twentysixteen#383 for background.

template-parts/content/content.php
```diff
@@ -31,16 +31,9 @@
...
-					array(
-						'span' => array(
-							'class' => array(),
-						),
-					)
-				),
-				get_the_title()
+				/* translators: %s: Name of current post. Only visible to screen readers */
+				esc_html__( 'Continue reading %s', 'twentynineteen' ),
+				'<span class="screen-reader-text">"' . get_the_title() . '"</span>'
```
ocean90: This change needs to be reverted to support proper translations in all languages. See https://core.trac.wordpress.org/ticket/31979 and WordPress/twentysixteen#383 for background.

mnelson4:
Also for reference: there is some current work that's going the opposite direction: replacing HTML with placeholders on https://core.trac.wordpress.org/ticket/45151#comment:15 (Josh found that one too, actually).

It seems there was previous concensus to use HTML instead of placeholders (of which I wasn't aware) in which HTML tags won 5-0, see https://wordpress.slack.com/messages/C02RP50LK, although admittedly the sample size was rather small, and I'm not aware of any p2 announcement being made about it.

So, I think folks like me could use some more education on the matter, probably in the form of docs. I'll try to bring this up with the docs team.

PR #285, opened 2018-10-24
Escape translated text

WPprodigy:
All output should be run through an escaping function.

WP.org Username: iCaleb

### DIFF
--- a/404.php
+++ b/404.php
@@ -17,11 +17,11 @@ get_header();
 
 			<section class="error-404 not-found">
 				<header class="page-header">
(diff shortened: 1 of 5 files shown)


ocean90:
Just like core, default themes don't escape translations when not necessary as those are trusted. See https://core.trac.wordpress.org/ticket/30724 for background.

WPprodigy:
@allancole So then should I open up a new PR that removes the current uses of these functions? 

- https://github.com/WordPress/twentynineteen/search?q=esc_html_e&unscoped_q=esc_html_e
- https://github.com/WordPress/twentynineteen/search?q=esc_html__&unscoped_q=esc_html__
- etc?

PR #286, opened 2018-10-24
Use the 3rd printf format arg in twentynineteen_posted_on()

WPprodigy:
Pretty minor issue. Just seemed odd to concatenate the `$time_string` while the rest are all using the printf format variables, especially while `$time_string` was already being passed as a replacement. 

WP.org Username: iCaleb


allancole:
Nice catch! Thanks for contributing ;-)

PR #290, opened 2018-10-24
Fix the alignment for the double arrow

Ismail-elkorchi:
The `$ui_icons` elements weren't aligned at the =>. 

Username: ismailelkorchi


allancole:
Looks good. Thanks for contributing! 👍

PR #292, opened 2018-10-25
Remove link attributes ‘aria-haspopup’ and ‘aria-expanded’

### LINKED ISSUE #270: Remove the `aria-expanded` attribute from the menu
The `aria-expanded` is a _state_, not a property, and it's meant to communicate a change in the state:

https://www.w3.org/TR/wai-aria-1.1/#aria-expanded
> Indicates whether the element, or another grouping element it controls, is currently expanded or collapsed.

As far as I see, in Twenty Nineteen `aria-expanded` is never updated and it's always `false`. As such, it gives an incorrect information to assistive technology users. For example, when a menu tree is open, it's still announced as "collapsed":

screen shot 2018-10-24 at 16 37 19

However, I'd suggest to remove `aria-expanded` entirely as it is better suited for UI controls that require _user activation_. As in:
- users activate an UI control
- `aria-expanded` communicates a change in the state (expanded, or collapsed)
- users are informed something expanded or collapsed in the page 
- they can now explore around for the new content or be aware some previously displayed content has been removed

Instead, the Twenty Nineteen menu doesn't require explicit user interaction to expand or collapse the sub-trees. It is sufficient to navigate normally pressing Tab or hover with the mouse to get the new content.

nielslange:
Fixes #270 

Before:

```
<a href="http://build.wordpress-develop.test/?page_id=2" aria-haspopup="true" aria-expanded="false">About The Tests</a>
```

After:

```
<a href="http://build.wordpress-develop.test/?page_id=2">About The Tests</a>
```

### DIFF
--- a/inc/template-functions.php
+++ b/inc/template-functions.php
@@ -184,27 +184,6 @@ function twentynineteen_get_discussion_data() {
 }
 
 /**
- * WCAG 2.0 Attributes for Dropdown Menus
- *
- * Adjustments to menu attributes tot support WCAG 2.0 recommendations
- * for flyout and dropdown menus.
- *
- * @ref https://www.w3.org/WAI/tutorials/menus/flyout/
- */
-function twentynineteen_nav_menu_link_attributes( $atts, $item, $args, $depth ) {
-
-	// Add [aria-haspopup] and [aria-expanded] to menu items that have children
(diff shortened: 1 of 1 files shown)


postphotos:
Reporting from WC Philly Contrib day. LGTM!

nielslange:
@allancole Can we 🚢 this PR?

PR #301, opened 2018-10-25
Fix for galleries column round 2

briannaorg:
Corrects gallery column alignment in issue #91.

Replace margin with padding so the width of flex-aligned gallery items
flexibly adjusts within its container. Set flex-flow settings to row
wrap.


kjellr:
This looks great. I tested it out and it works as it should. Would you mind clearing those two conflicts? Then this should be ready to merge!

kjellr:
Adding some screenshots for visual reference: 

_Before_
screen shot 2018-10-25 at 10 17 45 pm

_After_
screen shot 2018-10-25 at 10 17 27 pm

briannaorg:
Thanks @kjellr I appreciate it! I've resolved the conflicts and removed the .map file from the latest commit.

Issue #298, opened 2018-10-25
Fix post password field size/spacing issues.


kjellr:
Currently, the post password field is very large compared to the button next to it:

<img width="535" alt="screen shot 2018-10-24 at 10 00 48 pm" src="https://user-images.githubusercontent.com/1202812/47471477-69a30800-d7d8-11e8-8e94-e7f8e55774eb.png">

It should be the same height as the button, and should have a little extra space to the right and left:  

![password-field](https://user-images.githubusercontent.com/1202812/47471573-e33af600-d7d8-11e8-85ee-1694603e5ef8.png)

briannaorg:
Made some adjustments to the global input and button styles to match up vertically.

<img width="713" alt="screen shot 2018-10-24 at 10 35 03 pm" src="https://user-images.githubusercontent.com/3371846/47474837-922e0080-d7de-11e8-9f20-fbfa1a15b9fa.png">

kjellr:
Fixed via #304

PR #304, opened 2018-10-25
Align password input and submit button

briannaorg:
Fix for issue #298. 

Since there’s no specific styling for just this use case, and a fix
requires making changes to the global styles for inputs and buttons, I
just aligned the submit button type vertically and slightly adjusted
the padding so it’s equal to the height of all inputs, as well as
slightly adjusting the vertical padding of the inputs to match.

WordPress.org: @technosiren


briannaorg:
@kjellr no problem. style-editor.scss didn't have explicit padding on its button styles so I left it as-is. _blocks.scss has been updated to match _buttons.scss.

briannaorg:
@allancole - the vertical padding inside the inputs was reduced (shorter inputs), and the vertical padding of the buttons very slightly increased until they matched, after I vertically-aligned the submit button to the bottom of the input.  I tried to use a light touch on this one since it affects global styles, but @kjellr, let me know how you want to proceed.

kjellr:
> I assumed the input form height would be lowered to match the height of the button (instead of the other way around). I think I’d prefer shorter inputs over taller buttons, but I’ll let @kjellr make final call on this.

I'd originally assumed that'd be the most natural route too, but this actually feels pretty comfortable. It also turns out our global button sizes on the front end were unintentionally about 8px shorter than the buttons in the editor. 😄 So the changes in this PR not only fix the input field height, but they sort that out too, bringing greater alignment between the front and back end. Seems like a solid fix to me. 👍 

@briannaorg — would you mind fixing up that one conflict? I don't think `style.css.map` actually needs to be committed in this case. Once that's done we can go ahead and merge this.

briannaorg:
All set - I've merged master with this branch and omitted the .map file. Thanks @kjellr!

kjellr:
Thanks @briannaorg! I think the `.map` file is still included here... would you mind double checking that one?

briannaorg:
Thanks @kjellr - I am still getting used to the Github Desktop interface, the .map file has been deleted and committed to the branch.

Issue #282, opened 2018-10-24
Navigation sub-menu width broken after #263


afercia:
Seems to me the CSS property `max-content` introduced in https://github.com/WordPress/twentynineteen/pull/263 broke the layout of the sub-menus, see screenshot below (from Chrome on macOS):

![screen shot 2018-10-24 at 20 40 46](https://user-images.githubusercontent.com/1682452/47453657-99004780-d7cd-11e8-97aa-6996abf8d157.png)

I see the same on all browsers, on Windows too. Also, to my knowledge, `max-content` is not supported at all by Edge and IE11.

afercia:
Forgot to mention that any style applied on `:hover` should be applied also on `:focus` otherwise there's the risk the width of the sub-menu will break when using a keyboard to navigate.

kjellr:
Good catch. We'll need to fix. I'm not positive we need `max-content` there, but if we continue using it we'll need a fallback too.

nielslange:
@afercia Thanks for reporting this issue. On my screen, the implementation of #263 looked as followed:

![screenshot 2018-10-25 at 07 50 37](https://user-images.githubusercontent.com/3323310/47469379-cdeeb780-d82a-11e8-972d-9e63c632a75f.png)

Top left: Chrome 69.0
Bottom left: Opera 55.0
Top right: Safari 12.0
Bottom right: Firefox 64.0b3

Since I positioned the drop-down container under the parent menu item, I used the value `max-content` to ensure that the drop-down container uses its own width instead of the width of its parent element. I wonder if the following change would fix it:

```
width: 200px
max-width: max-content;
```

I do not have access to a Windows system and therefore cannot test this issue on Edge.

kjellr:
> I wonder if the following change would fix it:

It might — I'd have to take a look tomorrow. But I seem to remember that `max-content` in that context may also not have wide browser support. We'll have to double check on that. 

Also cc @c4rl0sbr4v0, who mentioned he'd take a look at fixing this.

afercia:
@nielslange thanks for clarification. It's perfectly possible I've made some mistake building the CSS. However, can you please try focusing a top level item with the keyboard? In that case, `max-content` doesn't get applied. Still not supported by Edge and IE.

kjellr:
This should be working as of #308. 👍

Issue #266, opened 2018-10-24
Main navigation menu not fully operable with a keyboard


afercia:
In Twenty Seventeen, it is possible to navigate the main menu with a keyboard tabbing in both directions. Pressing `tab` navigates through all the menu items, pressing `shift+tab` navigates through all the menu items in reverse order.

As far as I see, in Twenty Nineteen navigating forwards allows to go through all the menu items, including all the sub-items.

Instead, when navigating backwards, the sub-items are skipped.

westonruter:
Which browser(s) was this observed in?

afercia:
All of them. Really not a browsers issue.

westonruter:
I think I understand why this is happening: when tabbing forward you start with focus on the root link of the nav submenu. At this point, the links below are displayed and so they can then be tabbed to. However, if you tab past the last link in a submenu and then tab backwards, the last link in the submenu is no longer displayed and so it cannot be focused, meaning the keyboard jumps back to the previous closest focusable element: the root nav menu item.

afercia:
Yep. In CSS there's no selector to target an ancestor. The relatively new `focus-within` could be used to build a pure-CSS solution, see for example https://www.scottohara.me/blog/2017/05/14/focus-within.html. Unfortunately, `focus-within` is not supported by Edge and IE. I've heard there are polyfills but I've never played with them so I can't say if they're good. If they're good enough, it would be great to explore a pure-CSS solution.
Alternatively, Twenty Seventeen solves this issue with a little bit of JavaScript, I'd suggest to have a look at it.

afercia:
The ones I've found are:
https://github.com/jonathantneal/focus-within
https://github.com/jonathantneal/postcss-focus-within

westonruter:
I just found that the issue is not due to `:focus-within`. Rather, it's because the submenus get `display:none` when not displayed:

https://github.com/WordPress/twentynineteen/blob/3aa5fbee914e69002540fca20c160d56290ad509/sass/navigation/_menus.scss#L73

In Twenty Seventeen, however, these submeus get `display:block`:

https://github.com/WordPress/wordpress-develop/blob/5d92e0c0a950399c53294a41000305ddcf47efb5/src/wp-content/themes/twentyseventeen/style.css#L3515-L3519

For both Twenty Seventeen and Twenty Nineteen, the submenu that is not shown gets positioned off screen:

[...]

PR #308, opened 2018-10-25
Fix Issue #282 Regarding SubMenu Container width

c4rl0sbr4v0:
This PR fixes Issue: [Issue #282 ](https://github.com/WordPress/twentynineteen/issues/282)

I will let you know as soon I can test it on IE11, but a display table should fix it.
screenshot 2018-10-25 at 10 43 32

### DIFF
--- a/sass/navigation/_menus.scss
+++ b/sass/navigation/_menus.scss
@@ -74,7 +74,7 @@
 
 		display: none;
 		float: left;
-		width: 100%;
+		width: max-content;
 		position: absolute;
 		opacity: 0;
 		left: -999em;
@@ -133,12 +133,11 @@
 
 	.main-menu .menu-item-has-children:hover > .sub-menu,
 	.main-menu .menu-item-has-children .sub-menu:hover {
-		display: block;
+		display: table;
 		left: 0;
 		margin-top: 0;
 		opacity: 1;
 		position: absolute;
-		width: max-content;
 
 		.sub-menu {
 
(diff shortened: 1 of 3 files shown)


c4rl0sbr4v0:
This is how it is seen in IE 11
IE11

c4rl0sbr4v0:
Awesome! Didn't know if I should change or not that styling, that's why I didn't :-)

afercia:
I've noticed now that, when using a keyboard, the sub-menus appear thanks to `:focus-within` on their `.menu-item-has-children`.  Worth reminding `:focus-within` doesn't work on Edge and IE11 so when using the keyboard on those browsers the sub-menus can't be accessed in any way 😞 
/Cc @kjellr

kjellr:
Thanks, @afercia — since `:focus-within` was used prior to this PR, would you mind filing a separate issue for that one?

westonruter:
We should to add a polyfill for IE to toggle a `.focus-within` class similar to how core themes have done this .

Note that users who have JS disabled in their browsers would still have a keyboard-inaccessible nav menu, but this would have been the case for all core themes until now. With `:focus-within`, nav sub-menus are accessible for the first time for non-IE/Edge browsers which have JS disabled. So it's a both/and not an either/or situation.

afercia:
@kjellr sure, see https://github.com/WordPress/twentynineteen/issues/266#issuecomment-433308499 where coincidentally I've just suggested to try a polyfill.

PR #313, opened 2018-10-25
Automatically hide overflowing menu items in a dropdown menu

### LINKED ISSUE #268: Implement priority+ navigation pattern for main menu
As introduced in #32:

> ... a [priority+](
https://css-tricks.com/the-priority-navigation-pattern/) navigation pattern to help account for cases where there are very-long menus and menu items. This is also useful on mobile, where it becomes a slightly more efficient version of the usual ["hamburger"](
https://en.wikipedia.org/wiki/Hamburger_button) menu: rather than hide all menu items by default, it allows users to select some items directly without needing to tap first into a menu.

The priority navigation menu should keep the menu onto one line on both desktop and mobile screens. If there are items that don't fit on that single line, they should be hidden behind an ellipsis button, as shown below:

**Example of a long menu on desktop:**
47304568-02792e00-d5f5-11e8-9313-52ea6955f64d

**On small screens:**
47304572-04db8800-d5f5-11e8-9ae0-5fe3096e4b65

(Refer to #143 for some related menu work)

This will likely need to be written in JS, and preferably without the use of jQuery. A few possible implementation methods and examples are outlined here:

https://css-tricks.com/the-priority-navigation-pattern/

youknowriad:
closes #268 

This PR tries to add the automatic hiding of menu elements overflowing with the container. That way the menu always shows up in a single line.

In its current state, I did the minimal CSS changes to make it work, it’s still needs to be polished design-wise and probably a11y wise too but the behavior is there.

### DIFF
--- a/js/priority-menu.js
+++ b/js/priority-menu.js
@@ -1,0 +1,124 @@
+(function() {
+	/**
+	 * Prepends an element to a container.
+	 *
+	 * @param {Element} container
+	 * @param {Element} element
+	 */
+	function prependElement(container, element) {
+		if (container.firstChild) {
+			return container.insertBefore(element, container.firstChild);
+		} else {
+			return container.appendChild(element);
+		}
+	}
+
+	/**
+	 * Shows an element by adding a hidden className.
+	 *
+	 * @param {Element} element
+	 */
+	function showElement(element) {
+		// classList.remove is not supported in IE11
+		element.className = element.className.replace("is-hidden", "");
+	}
+
+	/**
+	 * Hides an element by removing the hidden className.
+	 *
+	 * @param {Element} element
+	 */
+	function hideElement(element) {
+		// classList.add is not supported in IE11
+		if (!element.classList.contains("is-hidden")) {
+			element.className += " is-hidden";
+		}
+	}
+
+	/**
+	 * Toggles the element visibility.
+	 *
+	 * @param {Element} element
+	 */
+	function toggleElementVisibility(element) {
+		if (element.classList.contains("is-hidden")) {
+			showElement(element);
+		} else {
+			hideElement(element);
+		}
+	}
+
+	var navContainer = document.querySelector(".main-navigation ");
+	// Adds the necessary UI to operate the menu.
+	navContainer.innerHTML +=
+		'<button class="main-menu-more is-hidden">More</button>' +
+		'<ul class="hidden-links is-hidden"></ul>';
+	var navContainer = document.querySelector(".main-navigation ");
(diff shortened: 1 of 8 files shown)


kjellr:
Thanks, @youknowriad! This looks like a great start. Either @allancole or I will jump in to start styling this today. 🙌 

Here's a GIF of the current state:  

menu

js/priority-menu.js
```diff
@@ -0,0 +1,121 @@
...
+			showElement(element);
+		} else {
+			hideElement(element);
+		}
+	}
+
+	var navContainer = document.querySelector('.main-navigation ');
+	// Adds the necessary UI to operate the menu.
+	navContainer.innerHTML +=
+		'<button class="main-menu-more is-hidden">More</button>' +
```
westonruter: How about adding this More button in PHP instead of JS? This would have the added benefit of facilitating translation of the label.

allancole: Yup, we’re working that out now.

youknowriad: My idea was to keep the php pure (avoid extra nodes) if the JS is not working for some reason. Your choice though

kjellr:
This still needs some work, but I pushed this a little closer. Some things that need to be sorted out before merge:

- [ ] In my testing, the priority+ menu didn't consistently kick in on every new page load. Sometimes I had to resize the window a little bit for it to activate. 
- [ ] For now, I exposed and indented all submenus in this iteration. I'm thinking we'll either need to replace that with the nested solution we have from #337 ([GIF here](https://github.com/WordPress/twentynineteen/pull/337#issuecomment-433585675)), or we'll need to adapt this method for all menu items. (It's simple and it saves users a tap or two).
- [ ] On mobile, you can currently scroll and see the normal page under the menu. 
- [ ] On mobile, we need to inject a close button.
- [ ] Our current desktop menus work on hover, but this requires a click. 
- [ ] On desktop, the only way to close this is to click the ellipsis button again. We should have it close whenever someone hovers outside of the menu (Or clicks, depending on whether or not we use a click or a hover to trigger the menu, as per the above bullet point). As it stands now, you can activate both a submenu _and_ the priority+ menu, which leads to [some messiness](https://cloudup.com/cv-Oi-JbwCr). 
- [ ] On desktop, the menu is just anchored to the right of the header area. This is usually okay, but it'd be better if this was more closely aligned with the button itself. 
- [ ] This should be tested more. 😄 

Here's the current state in GIF form:

_Large screens:_
priority-plus

_Small screens:_
priority-plus-mobile

Issue #223, opened 2018-10-20
Fixed product layout broken in search result archives page.


mmaumio:
Create few products and search with a keyword to see the product post layout broken.
Screenshot added for reference.
![screenshot 2018-10-21 at 4 42 36 am](https://user-images.githubusercontent.com/3687038/47261210-1dac4680-d4ec-11e8-94db-d72087ffb1c4.png)

Fixed here: #222

joyously:
Is this just because the margin is not applied for unknown post types?

allancole:
I think we’ll end up adding an `.entry` class to _all_ post types to deal with this issue and the concerns mentioned here: https://github.com/WordPress/twentynineteen/issues/74

mmaumio:
@allancole Would you like me to try this? 
What I have understood is, we'll remove the hentry class and push an entry class to the post class of every posts regardless of the product or any other custom post type.
Then write the css of "hentry" to the "entry" class.

Is that correct?

Thank You

joyously:
> we'll remove the hentry class

Don't add or remove the `hentry` class. _(Well, except that it shouldn't be on Pages, but that might be a different issue.)_ 
Add `entry` class to all posts.
Change CSS that refers to `hentry` to instead use `entry`.

allancole:
@mmaumio, @joyously is mostly correct. 

To clarify:

• Don't add, change or remove the `.hentry` class.
• Use the `add_filter( 'post_class', '...' )` filter to add the `.entry` class to _all_ post-types.
• Change any CSS selectors that use `.hentry` to use `.entry`, instead.

mmaumio:
That's great.
I'm on it.

Thanks

mmaumio:
Made the changes recommended.

WPprodigy:
Linking to the PR for search-ability: https://github.com/WordPress/twentynineteen/pull/229

allancole:
@mmaumio This should be fixed in #316 so I’ll close this for now. If you still see the problem feel free to reopen or start a new issue.

mmaumio:
Sure.
I'll let you know.

Thank You

PR #316, opened 2018-10-25
Replace .hentry class in header for better code quality

allancole:
Replaces the `.hentry` class in the header with `.site-featured-image` and. 

Applies the issues mentioned in #229 and #223


joyously:
But in this usage, is it in the right place? It is added by core from the `post_class()`. Is that used here?

kjellr:
To clarify, this PR did not remove the main `hentry`. When there was a featured image on a single post page, we unnecessarily had two separate areas with the `hentry` class. This removes the incorrect, manually-added one.

PR #317, opened 2018-10-25
Remove unnecessary use of alt text in image ::after/::before

### LINKED ISSUE #65: Untranslatable text in CSS

audrasjb:
Fixes #65 and closes former PR #99 
Removes unnecessary alt text (which was even not translatable) from ::after CSS/SASS rule.

Screenshots of the issue:
47058246-921d7780-d1c4-11e8-8e5b-2d52b791aaa1


allancole:
Looks good! Just needed to resolve some conflicts with `master` for you. Thx!

Issue #62, opened 2018-10-16
Font sizes should be ems or %, never px


joyously:
https://github.com/WordPress/twentynineteen/blob/7b1170c700b51ccc89479a5552c97102832b2d77/style.css#L395

kjellr:
I generally tend to agree, however [Gutenberg does use `px` font sizes](https://github.com/WordPress/gutenberg/blob/286317c8e61b1a2922243875fe1f98c868d7859b/assets/stylesheets/_variables.scss), and we've adopted them here in order to maintain 1:1 font size consistency in the front and backend. 

(If someone has another solution, I'd welcome a PR!)

joyously:
If the editor inserts styles for font sizes, that has nothing to do with the theme.
The line quoted is for the `html` tag. This should be 100%, so that the user gets the font size he chose in his browser.
But other references to pixel sizes should not be used. Really, there is not much need to set font sizes except for headings as ems, and `<big>` and `<small>` which should be %.

2ndkauboy:
As `em` and `%` are both relative units, they need a value to be relative to. So using a fixed `font-size` in `px` is usually the way to go. Having the fixed size (only) on the `html` tag is good practice.

The size of the text can easily be increased by the browser, even if defined in `px`, which is very important for the a11y. Setting the `font-size` to `100%` would result in a size of `16px` in "most" browsers, but is not a reliable value to use relative units from.

joyously:
> Setting the font-size to 100% would result in a size of 16px in "most" browsers

Yes, this is what is needed, because those that have changed their default size have a reason to do so. Setting it to a fixed px size means that you override what the user chose. This is a concern for accessibility, for those that have low visual acuity and set their default sizes to larger. Just this weekend, I was helping an elderly man who had problems with his retinas, and had to set his browser font size up to quite large to see well. By setting it to 22px, it would ignore his setting, and he couldn't read it.

2ndkauboy:
The `font-size` in the browser is usually scale "relative" and not absolut to a different value, so setting the size to `22px` it also perfectly fine, if a user zooms. Also the browser zoom (using CTRL and the mouse wheel or `+`) is also zooming relative.

2ndkauboy:
As an alternative to just having `px`, we could use it together wirh `rem`:

```css
html {
    font-size: 22px;
    font-size: 1.375rem;
}
```

joyously:
[...]

PR #318, opened 2018-10-25
Fix font size variables to resolve to exact pixels

### LINKED ISSUE #164: Font size variables should resolve to exact pixels.
Currently, most of our font size variables (all set in `em`) don't equate to nicely rounded pixel sizes:

| Variable  | Pixel Size  |
|---|---|
| `$font__size-xs`  |  `15.644px` |
|  `$font__size-sm` |  `19.555px` |
| `$font__size-md`  | `22px`  |
| `$font__size-lg`  | `37.125px`  |
| `$font__size-xl`  | `49.5px`  |
| `$font__size-xxl`  | `61.875px`  |

This can lead to messy half-pixel rendering — we should revise our type scale so that it results in exact, crisp, pixel-perfect text. 💎 🎉

2ndkauboy:
Fixes #164

Username: Kau-Boy

### DIFF
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -199,7 +199,7 @@
 		li {
 			color: $color__text-light;
 			font-family: $font__heading;
-			font-size: calc(#{$font__size_base} * #{$font__size-lg / 1em} );
+			font-size: calc(#{$font__size_base} * #{$font__size-lg} );
 			font-weight: bold;
 			line-height: $font__line-height-heading;
 			
--- a/sass/site/secondary/_widgets.scss
+++ b/sass/site/secondary/_widgets.scss
@@ -28,7 +28,7 @@
 		li {
 			color: $color__text-light;
 			font-family: $font__heading;
-			font-size: calc(#{$font__size_base} * #{$font__size-lg / 1em} );
+			font-size: calc(#{$font__size_base} * #{$font__size-lg} );
 			font-weight: bold;
 			line-height: $font__line-height-heading;
 
--- a/sass/typography/_typography.scss
+++ b/sass/typography/_typography.scss
@@ -1,6 +1,6 @@
 
 html {
-	font-size: $font__size_base;
+	font-size: 100%;
 }
 
 body {
@@ -9,7 +9,7 @@ body {
 	color: $color__text-main;
 	font-family: $font__body;
 	font-weight: 400;
-	font-size: 1em;
+	font-size: $font__size_base;
 	line-height: $font__line-height-body;
 	margin: 0;
 	text-rendering: optimizeLegibility;
--- a/sass/variables-site/_fonts.scss
+++ b/sass/variables-site/_fonts.scss
@@ -5,17 +5,16 @@ $font__heading: -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen
 $font__code: Menlo, monaco, Consolas, Lucida Console, monospace;
 $font__pre: "Courier 10 Pitch", Courier, monospace;
 
-$font__size_base: 22px;
-$font__size-ratio: 1.125;
+$font__size_base: 1.375rem;
 
-$font__size-xxs:   1em / (1.5 * $font__size-ratio);
-$font__size-xs:    1em / (1.25 * $font__size-ratio);
-$font__size-sm:    1em / (1 * $font__size-ratio);
-$font__size-md:    1em * (1 * $font__size-ratio);
-$font__size-lg:    1em * (1.5 * $font__size-ratio);
-$font__size-xl:    1em * (2 * $font__size-ratio);
-$font__size-xxl:   1em * (2.5 * $font__size-ratio);
-$font__size-xxxl:  1em * (3 * $font__size-ratio);
+$font__size-xxs:   0.875rem;
+$font__size-xs:    1rem;
+$font__size-sm:    1.25rem;
+$font__size-md:    1.375rem;
+$font__size-lg:    1.75rem;
+$font__size-xl:    3rem;
+$font__size-xxl:   4rem;
+$font__size-xxxl:  4.5rem;
 
 $font__line-height-body: 1.8;
 $font__line-height-pre: 1.6;
--- a/style-editor-frame.css
+++ b/style-editor-frame.css
@@ -18,7 +18,7 @@ body.block-editor-page .block-editor .editor-post-title__block:before {
 }
 
 body.block-editor-page .block-editor .editor-post-title__block:before {
-  width: 2.8125em;
+  width: 4rem;
   margin-top: 0;
   margin-bottom: 0;
   margin-left: 1em;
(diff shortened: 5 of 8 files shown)


kjellr:
Interesting! This seems promising. 

Question for you:  From what I can tell, the new font size variables here _should_  equate to exact pixel values (as shown in this [CodePen](https://codepen.io/kjellr/pen/ePbazY)). But they're not doing that for us. Is that because of rounding/truncating of decimal points in our Sass compiler? If so, I wonder if we can tweak that. If another few decimal places will mean we have more exact pixel sizes, I'm all for it.  

Regarding the new type scale: The new font sizes here sound reasonable, but I'll likely want to give this a more in-depth review over the weekend. I know we'd been using ~37px text in places that will be either much larger or  much smaller with this new scale. Also, this PR helped me notice that we have an unintended ~`25px` font size used in many areas of the site (the site header for example). When we switch to this PR, that text is sized down to 22px. 😄 In general, I may want to make some slight adjustments to the font sizes to smooth these things out. 

In any case, I'll take another in-depth look at this over the next couple days. I'll also encourage @allancole to take a look too.

Thanks again for the PR! This is really helpful.

2ndkauboy:
In `em`, many values cannot round to exact `px` values, given a base of `22px`. If we would use a base on the `html` of `10px` we could set the base for the `body` to `2.2em` and all other values would perfectly round to values such as `4.8em` instead of `2. 18182` with the base value `22px`.

kjellr:
> In em, many values cannot round to exact px values, given a base of 22px. If we would use a base on the html of 10px we could set the base for the body to 2.2em and all other values would perfectly round to values such as 4.8em instead of 2. 18182 with the base value 22px.

Right, right. I guess I was just puzzled why the CodePen _does_ produce exact pixel values for me, whereas this doesn't. It must be the rounding. CodePen equates `$font__size-lg` to be  `1.2727272727em`, while our CSS compiler outputs `1.27273em`. 

As you mention, we can't do much better in this case, short of assigning the base font size to be something nice and round like `10px` (which I _think_ has the potential to cause issues somewhere, right?)

2ndkauboy:
I am currently at a conference, but I try to create a Code Pen solving the issue of rounded values, which you also take the issue #62 into account.

2ndkauboy:
@kjellr: This could be an solution using rem with a `1em` = `10px` on the `html` tag using font-size: 62.5%`: https://codepen.io/anon/pen/qJvrNL

To support older browsers, we could use a simple "rem to px" PostCSS script, like this one: https://github.com/robwierzbowski/node-pixrem

joyously:
> ` font-size: 62.5%`

Before you do that, _please_ reread what I wrote in the other font ticket about using 100%.
And I've seen a lot of themes that use Bootstrap with that in there and then they have issues with nested things being tiny (because they use `rem` wrong).

2ndkauboy:
I am sorry @joyously but I don't see the issue here. Using `rem` (and not `em` will not result in things being tiny when nested. The `body` still has the default size of `22px` equivalent. I can resize the default font in Chrome using both optione (zoom the whole page or only setting the font-size to another browser default). Everything just resizes perfectly. I also can't see any difference, when using 100% for the `html`. Can you give us an example, where these definitions would fail?

joyously:
I never said they would fail. I said please don't use anything but 100% (or larger) on html. It is still an accessibility issue, and it should not be used to get pixel division correct. 
In an ideal world, pixels would never be used for fonts, because you don't know what your base is. Giving the user easier-to-use controls for font size in the editor opens a Pandora's box.. It really is not such a good thing. 

The examples I have seen, reviewing many themes submitted to the repository, is that Bootstrap uses the 62.5% rule and then there is something (maybe in _s) that uses rem. A lot of authors don't test everything, so they don't see it, but I would see it in my test data in the same place every time (some nested thing that should have used em, but had rem). I dislike rem, because everything's default is already working fine. Nothing really needs a font size change, and when you change it using rem it is difficult to make it work for all nested things. If a font size change is needed, it should be a class on the container, to affect children equally, using percent or ems. That way, there are no nesting font size problems. (I know `<big>` and `<small>` should be on the elements, using %.)

2ndkauboy:
I still don't see the a11y issue here. The combination of `62.5%` on the `html` and `2.2rem` on the `body` will result in `100%` of the users default size (which is usually `16px` for most browser).

From that size, a designer can use `em`, `%`, and other relative sizes and they would all scale the font without any a11y issue. I couldn't find a single blog post about font-sizes and a11y where a base size of something different than `100%` is an issue. They all just say, that the font-size hasto be resizable.

If we do want to use `22px` as the default size of the `body` font-size (no matter in which unit we accompish that), useing `100%` on the `html` would not make it possible to have all the other sizes rounded to exact pixel values.

2ndkauboy:
Oh, do you acutally want to say, that we should not use a font-size of `22px` but of `16px` (`100%`)? Then I know were the miscommunication might come from :)

2ndkauboy:
OK, maybe I found a good solution: https://codepen.io/anon/pen/NOJaxY

The `html` font-size is at `100%`, the `body` font-size is at `22px` and all the other values round perfectly without any rounded periodic numbers.

joyously:
Your codepen is better, but not quite there. There is no need for a rule for the body, and by having one, it sort of eliminates the value of using 100% on html.
Why would you make the base size equivalent to extra small? Shouldn't the base be in the middle, and the user could make smaller or larger text? (I see that you made medium the same number as base, but body should not have base, so it would be extra small.)

Using rems on these sizes makes them set that size, regardless of nesting, but if anything with a % or ems is used to change the whole container, these will not change with them. Consider blockquote with nested text that has one of these. (Of course, you'd have the same problem if you tried to use pixels.)

kjellr:
Thanks, @2ndkauboy! Your latest CodePen looks like a definite improvement over what we have. _Much_ cleaner than the current solution. Mind spinning up a PR with that so we can try it out in context? I think it'll work fine, but I want to make sure these correctly override some of Gutenberg's font size values. 

Thanks for continuing to push this! Your work is really appreciated. 🙌

joyously:
Hey, they are doing a contributor day in Philadelphia and going through all the Gutenbugs. One mentioned was https://github.com/WordPress/gutenberg/issues/11043 which is saying that it doesn't work...

2ndkauboy:
@kjellr: As there have been a lot of commit since opening this PR and it was conflicted, I updated it with the code from the latest CodePen.

kjellr:
@2ndkauboy, just wanted to let you know I haven’t forgotten about this one. This change will likely have some cascading effects on font styles throughout the site (I’m specifically thinking about the editor styles), so I doubt we’ll be able to get it in by RC1. Especially since we’ll need to continue focusing on finding/fixing bugs for that release. 

Although this is lower priority, I do still have hope we’ll get this tested and updated somehow. Keeping my fingers crossed!

2ndkauboy:
OK, no problem. Should I fix the conflict or wait until we proceed with this issue?

kjellr:
Let's hold off for now — I think fixing existing issues and general testing is more important at the moment. 👍 Thanks!

kjellr:
Hi @2ndkauboy! Since this PR is stale, I'm not going to migrate it over to Trac alongside the new ticket: 

https://core.trac.wordpress.org/ticket/45899 

Thanks for your work on this, and I hope we can get it sorted out in the future.

Issue #296, opened 2018-10-25
Fix pullquote citation text size. 


kjellr:
On the front end, It looks like the `cite` tag inside of pullquote blocks is now nested inside of a `p` tag (as of Gutenberg 4.1.0). This causes our `em`-based font sizes to nest, resulting in larger than intended citations: 

<img width="472" alt="screen shot 2018-10-24 at 9 38 34 pm" src="https://user-images.githubusercontent.com/1202812/47470836-812cc180-d7d5-11e8-916f-cdf7eb455b55.png">

The text size there should match the text size in the back end:

<img width="490" alt="screen shot 2018-10-24 at 9 41 39 pm" src="https://user-images.githubusercontent.com/1202812/47470855-a0c3ea00-d7d5-11e8-8d24-4f2ececcfbe5.png">

kjellr:
Automatically resolved in Gutenberg 4.1.1.

PR #319, opened 2018-10-25
Correct nested citation font size in pullquote block

briannaorg:
Fix for #296

I switched the font size calculation so it’s using the global font size
base instead of inheriting it from the parent paragraph tag.

WordPress.org: @technosiren

Frontend:
screen shot 2018-10-25 at 10 09 54 am

Editor:
screen shot 2018-10-25 at 10 09 46 am


kjellr:
Thanks @briannaorg — oddly enough, I'm not sure this is needed anymore. 😄 I just updated to Gutenberg version 4.1.1 and it's no longer wrapping these in a  `p` tag. Can you test that out and confirm?

briannaorg:
Just confirmed it's no longer wrapping cite tags in p tags in block output in Gutenberg 4.1.1!

My one argument for keeping this PR is that it removes the inheritance of <cite> font size from any parent container, which basically future proofs as Gutenberg gets finalized. No worries either way!

kjellr:
Thanks, @briannaorg!

> My one argument for keeping this PR is that it removes the inheritance of font size from any parent container, which basically future proofs as Gutenberg gets finalized. No worries either way!

I'm going to close it for now, since it would introduce a potential non-standard font size in the future (one that's not directly tied to one of our font size variables). This was an okay tradeoff if it was necessary here, but I'd like to avoid it if we don't need it right now. 👍

Issue #284, opened 2018-10-24
Allow users to turn off image filters


kjellr:
The theme already includes a filter to turn off our featured image filters: 

https://github.com/WordPress/twentynineteen/blob/7eb93d5f79eedb168f72ccfeddbfcf6f0f129875/inc/template-functions.php#L120

We should make this a theme option, so it's available within the Customizer. If a user chooses to turn off image filters:

- Featured images within archive pages should have no overlay on them at all.
- Featured images on single post pages have text on top of them, so they should have an overlay. I suggest adding a layer of `black` at `0.7` opacity in that case:

### Screenshots, with filters disabled:

_In post feed, archive pages (no overlay):_

![screen shot 2018-10-24 at 4 22 47 pm](https://user-images.githubusercontent.com/1202812/47459070-31d09c00-d7a9-11e8-8d76-bb807eb8a1da.png)

_On single post pages (black overlay at `0.7` opacity):_

![screen shot 2018-10-24 at 4 15 13 pm](https://user-images.githubusercontent.com/1202812/47459107-46ad2f80-d7a9-11e8-903e-f376a20300d8.png)

nielslange:
Closed as fixed when #327 will be merged.

PR #320, opened 2018-10-25
fix for opacity featured image

LittleBigThing:
Proposed fix for the opacity of featured image layers for browser not supporting mix-blend-mode
Looks good on Chrome and Safari, need to test on Edge and IE.

### DIFF
--- a/sass/mixins/_mixins-master.scss
+++ b/sass/mixins/_mixins-master.scss
@@ -39,9 +39,13 @@
 
 	&:after {
 		background: $color__link;
-		mix-blend-mode: multiply;
-		opacity: 1;
+		opacity: 0.8;
 		z-index: 3;
+		
+		@supports (mix-blend-mode: multiply) {
+			mix-blend-mode: multiply;
+			opacity: 1;
+		}
 	}
 }
 
--- a/sass/mixins/_mixins-master.scss
+++ b/sass/mixins/_mixins-master.scss
@@ -42,6 +42,7 @@
 		opacity: 0.8;
 		z-index: 3;
 		
+		// Browsers supporting mix-blend-mode don't need opacity < 1
 		@supports (mix-blend-mode: multiply) {
 			mix-blend-mode: multiply;
 			opacity: 1;
--- a/sass/site/header/_site-featured-image.scss
+++ b/sass/site/header/_site-featured-image.scss
@@ -139,6 +139,7 @@
 		opacity: 0.8;
 		z-index: 3;
 		
+		// Browsers supporting mix-blend-mode don't need opacity < 1
 		@supports (mix-blend-mode: multiply) {
 			mix-blend-mode: multiply;
 			opacity: 1;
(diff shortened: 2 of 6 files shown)


LittleBigThing:
Should fix #121

LittleBigThing:
Will add some comments later

LittleBigThing:
Not sure about the conflicts, but the PR works as expected in IE and Edge (#121). Of course, the opacity can always be adjusted as needed. Accessibility testing is welcome. Also, some opinions about the effect.

Personally, I like the single post version with the image as a background. The featured image in the post feed is trickier: there, the image is not meant to be a background image, so one could expect to see it more clearly. This is of course general comment, but is more pronounced for this fallback version as the effect is less clear and unique.

Note: `filter` is not supported in IE11 so the result is different there than is Edge.

kjellr:
This is looking great, thanks @LittleBigThing! 

Here are some screenshots of my testing:

**Safari 12.0:**

screen shot 2018-10-25 at 10 34 42 pm
screen shot 2018-10-25 at 10 34 09 pm

**Edge**

screen shot 2018-10-25 at 10 37 43 pm
screen shot 2018-10-25 at 10 37 07 pm

**Internet Explorer**

screen shot 2018-10-25 at 10 39 22 pm
screen shot 2018-10-25 at 10 39 07 pm

---

Any idea why the post-feed versions for IE/Edge are so much lighter than the ones on the single post page? Are those missing a layer or something?

> The featured image in the post feed is trickier: there, the image is not meant to be a background image, so one could expect to see it more clearly. This is of course general comment, but is more pronounced for this fallback version as the effect is less clear and unique.

Yeah, I agree. It's _a little_ weird, but it doesn't bother me too much. It's also worth noting that site authors will have the option to opt-out of these overlays entirely in case it's a problem for them (#284). 

As per the notes in #6, this should be fairly solid on the a11y front. The lightest color possible with these alternate overlays is actually darker than that of our usual overlays (which already passes WCAG AA). 👍 

Once we sort out those merge conflicts (I'm happy to take a stab at it if you'd like — just let me know), and figure out why there's a lightness difference between the in-feed images and the single-post images, this should be good to go. 

Thanks again for seeing this issue through! 🙌

LittleBigThing:
I am not sure why it looks the same on browsers that support `mix-blend-mode`, probably something to do with the effect it generates, but indeed, the post-feed version could use the additional dark layer on IE and Edge (the fourth layer of the featured image styles on single posts). Something like this:

```
.post-thumbnail-inner:after {
    background: rgba(0,0,0,.35);
    content: "";
    display: block;
    height: 100%;
    opacity: .5;
    position: absolute;
    top: 0;
    width: 100%;

    @supports (mix-blend-mode) {
        display: none;
    }
}
```

LittleBigThing:
Far from being a Github expert, I suppose that I better start all over to fix those conflicts over after merging the option of turning off the image filter (#284)? :-)

kjellr:
> Far from being a Github expert, I suppose that I better start all over to fix those conflicts over after the option of turning off the image filter will be merged (#327)? Or is there a better order to merge? :-)

I don't _think_ there will be issues, but I do think that #327 will be ready to go shortly, so let's hold off just to double check. 👍

kjellr:
Closing in favor of #375.

Issue #277, opened 2018-10-24
Links in the content don't have a focus style


afercia:
As far as I see, links within the post content don't have any focus style. Indication of focus must always be clear and predictable for all users.

00travelgirl00:
Reladet PR #321

I could not find out how to style just the links in entry-content with sass.

kjellr:
Fixed via #321. 

<img width="764" alt="screen shot 2018-10-26 at 10 52 35 am" src="https://user-images.githubusercontent.com/1202812/47574294-45951300-d90d-11e8-9f5b-c00c7e7c4fd5.png">

PR #321, opened 2018-10-25
add outline style to focus to make active links visible

00travelgirl00:
Currently the a:focus for entry-content does not have an visible inpact. 

Therefore I suggest to add an outline (thin & dotted) to show the link. 

links_no_visible_focus

WordPress.org Username: travel_girl  :-)


00travelgirl00:
reladet Issue: #277

PR #311, opened 2018-10-25
Fixed avtar overlap issue

iamchetanp:

PR #324, opened 2018-10-25
Fix overlapping Gravatar issue and remove avatar for logged out users

allancole:
Building on the work from #214 and #311, this makes the avatar visible for logged-in users on all screen sizes while also removing the avatar entirely for logged-out users. 

See here: https://github.com/WordPress/twentynineteen/issues/213#issuecomment-431587683

It also removes some unused css.


inc/template-tags.php
```diff
@@ -219,16 +219,8 @@ function twentynineteen_discussion_avatars_list( $comment_authors ) {
 	function twentynineteen_comment_form( $order ) {
 		if ( true === $order || strtolower( $order ) === strtolower( get_option( 'comment_order', 'asc' ) ) ) {
 
-			/* Only show avatar for logged in users */
```
allancole: Removing the avatar altogether here. It’s only there as a design decision but isn’t very useful.

Issue #191, opened 2018-10-19
Make the primary site color customizable. 


kjellr:
Currently, our primary blue color (`#0073aa`) is hardcoded into the stylesheet. We need to give people the opportunity to customize it. 

Related PRs: 
- #113: Adding basic color selection for featured images.
- #149: Expanding the color selection to every instance of the blue.

fabiankaegy:
Just continued working on this. One question regarding @allancole's answer yesterday inside #149 is still open. Should there be separate color pickers for each element or just the primary color and the hover state for links buttons?

designsimply:
Closed #85 and #149 in favor of consolidating here. Noting that opacity is covered separately in #121.

celloexpressions:
One of the challenges with custom theme colors historically is ensuring that colors have adequate contrast. [The plugin](https://wordpress.org/plugins/fourteen-colors/) that allows color customization for Twenty Fourteen uses [this approach](https://celloexpressions.com/blog/a-strategy-for-custom-colors-in-the-customizer/) to automatically check and adjust contrast. Twenty Seventeen uses a hue picker and  hardcoded saturation and lightness values in hsl-defined colors to maximize flexibility while maintaining the general design intent.

For the featured image color, a hue picker may make more sense than a color picker. This offers a more-controlled customization that minimizes potential color contrast issues and maintains the design intent. A full color picker could make sense for the other uses of the blue, potentially with built-in contrast adjustments. I do recommend discussing all proposed color customization in one place as the implementation should be integrated in both code and UI. We should consider that the WordPress Philosophy includes several points that would encourage the use of fewer distinct options, which manage more colors with automated adjustments.

The Twenty Fourteen approach (which was ultimately shipped as a plugin due to the release schedule) is documented with some modernization of the customize API here: https://celloexpressions.com/blog/a-strategy-for-custom-colors-in-the-customizer/. This includes a library for color contrast checks; we would also want to leverage the customizer's [notifications API added in WordPress 4.6](https://make.wordpress.org/core/2016/07/05/customizer-apis-in-4-6-for-setting-validation-and-notifications/).

I haven't put together a post detailing the Twenty Seventeen approach yet, but would be happy to do so if that's helpful. I can definitely help review any color-customization-related pull requests if you ping me. Twenty Seventeen also included a "dark" mode, which Twenty Nineteen could be a good candidate for as well. Here are a couple of Twenty Seventeen links for reference:
[...]

PR #327, opened 2018-10-25
Customizer: introduce theme option to disable the image filter

### LINKED ISSUE #284: Allow users to turn off image filters

eliorivero:
Fixes #284 

This PR introduces a new section in Customizer with a single option to activate or deactivate the filter applied to featured images.

### Customizer control

captura de pantalla 2018-10-27 a la s 13 59 33

### Black overlay on single view

Additionally, @kjellr requested in the issue that a black overlay be added to the featured image in single post view. This is how it looks

captura de pantalla 2018-10-25 a la s 14 47 20

### Testing instructions

Checkout this branch and build the updated styles with the command line:
```
npm install && npm run build:style
```
Now you can go to the Customizer and try the new control.


LittleBigThing:
Cool! :-)

Maybe a detail: shouldn't the control be a checkbox?
Ideally, the control to select the custom color for the filter should be put below this one and displayed conditionally when this control is active. So, they would need to be in the same section.

kjellr:
Very cool, thanks @eliorivero! This works well. Here's a GIF of this in action:

image-filter

Just a couple procedural details to figure out...

---

I have a rough rewrite of the copy:

> **Featured Image Color Filter **
> By default, Twenty Nineteen adds a duotone-like effect to featured images using your site's primary color. This effect can be turned off. When the color filter is disabled, a standard black overlay will be used on single post pages to preserve readability of the text that sits on top of the featured image.
>
> Featured images should:
> [] Have a color filter applied.
> [] Not have a color filter applied.

I'd love to see if @michelleweber has some copy thoughts though  — she's always great with this sort of copy. 🙂

---

> shouldn't the control be a checkbox?

I like the radio buttons in this case, because I tend to think of checkboxes as controls that require a second action (like a confirm/submit) button to set in action. That said, I was just clicking around the Customizer and I can see that the "Display Site Title and Tagline" option is a checkbox that shows its impact immediately. 🙂 I'd like to loop in @melchoyce for a quick checkbox vs. radio button weigh-in too. 

---

> Ideally, the control to select the custom color for the filter should be put below this one and displayed conditionally when this control is active. So, they would need to be in the same section.

The custom filter color is actually planned to replace the primary color for your site (basically, everything blue right now: links, buttons, etc.), so it makes more sense to keep that option separate, in the "Colors" section.

kjellr:
> Removing the "that sits" solves the orphan, not sure if it's acceptable wording

Thanks @eliorivero — that works. 👍

kristastevens:
Thanks for the ping! I have some suggestions to consider within this one sentence: 

Original: 
> When the color filter is disabled, a standard black overlay will be used on single post pages to preserve readability of the text that sits on top of the featured image.

Proposed: 
> When the color filter is disabled, a standard black overlay appears on individual post pages to preserve readability of the text on top of the featured image.

allancole:
> Ideally, the control to select the custom color for the filter should be put below this one and displayed conditionally when this control is active. So, they would need to be in the same section.

Thought this through a bit and I think @eliorivero had it right here ^^. The Image Filter option is ultimately a decision about color and I think users will have an easier time discovering it if it appears on the same Customizer panel as the Color options. If you add this option to the same panel as #191 _and_ make it appear conditionally, I think we can go ahead and get this in ;-)

kjellr:
> When the color filter is disabled, a standard black overlay appears on individual post pages to preserve readability of the text on top of the featured image.

Thanks, @kristastevens! @eliorivero: let's update that sentence. 

> If you add this option to the same panel as #191 and make it appear conditionally, I think we can go ahead and get this in ;-)

Sounds good to me too. If this is ready before #191, I suggest we don't worry about the conditional nature yet. We can just move this over to the Colors section and merge it in.

melchoyce:
Fun idea!

This is likely scope creep, but what if we did something like:

( O ) Color Filter
[#0000ff] Color Picker
(     ) Black Filter 

That would make it an option between color or black, rather than something turned on/off. I also think it would be rad to let people choose the color overlay.

kjellr:
Looks good, thanks @eliorivero! Can we make one minor change? We should add a "Featured images should:" label above the button options. 

> ( O ) Color Filter
[#0000ff] Color Picker
( ) Black Filter

I do like that, but it'd work only if we let people choose a custom color there in addition to their site's primary color. So let's hold off for now and get this initial functionality in place.

eliorivero:
I've added the label. Please note that we already used the label and the description for this control, so I'm faking it at the end of the description with some CSS. Here's the screenshot (updated in the PR description too) and the previous iteration without the label.

captura de pantalla 2018-10-26 a la s 14 51 18captura de pantalla 2018-10-26 a la s 13 22 23

 I could've made the font bigger like the control's real label but I'm afraid to conflict with the typography hierarchy. This is how it looks when the font size is the same than the label. What do you think?

captura de pantalla 2018-10-26 a la s 14 58 40

melchoyce:
I still think those options are a little confusing — "not have a color filter applied" sounds like you won't necessarily have _any_ filter applied, and I'm sure many people will gloss over the description.

What about:
> *Featured images will*:
> Have a color filter applied.
> Have a black filter applied.

kjellr:
> Featured images will:
> Have a color filter applied.
> Have a black filter applied.

@melchoyce You're right, that's more clear. 🙌 

@eliorivero, if you can slip that in too that'd be awesome. 🙂

kjellr:
> The second label makes it look like the black filter will be applied everywhere but the featured image doesn't have any color filter in the post list. 

Thanks @eliorivero. I chatted with @benhuberman about that point, and we both felt it's okay, as long as we spell things out clearly in the description. He also had some thoughts on how to trim down and clarify the copy just a bit: 

> *Featured Image Color Filter*
>Twenty Nineteen adds a color filter to featured images using your site's primary color. If you disable this effect, the theme will use a black filter in individual posts to keep text readable when it appears on top of the featured image.
>
>On Featured Images, apply
> A color filter
> A black filter

eliorivero:
Done, all updated 👍

eliorivero:
Awesome, thanks for all the reviews and the merge 👍

PR #336, opened 2018-10-26
Style pingbacks

### LINKED ISSUE #299: Style pingbacks
Currently pingbacks are completely unstyled. These should have some sort of treatment. 

screen shot 2018-10-24 at 10 03 12 pm

nielslange:
Fixes #299 

<table>
<tr>
<td>Before:

299 - before</td>
<td>After:

299 - after</td>
</tr>
</table>

### DIFF
--- a/sass/site/primary/_comments.scss
+++ b/sass/site/primary/_comments.scss
@@ -107,18 +107,26 @@
 	}
 
 	.pingback {
-		padding-left: calc(.5 * (#{$size__spacing-unit} + calc(100vw / 12 )));
-		&:first-child:before {
-			content: 'Pingback:';
+
+		.comment-body {
 			color: $color__text-light;
 			font-family: $font__heading;
-			font-size: $font__size-sm;
-		}
-		.comment-body {
-			font-size: 0;
-			a {
-				font-size: $font__size_base;
+			font-size: $font__size-xs;
+			font-weight: 500;
+			margin-bottom: $size__spacing-unit;
+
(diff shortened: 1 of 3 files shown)


allancole:
@kjellr what do think about borrowing and modifying the style from the “Categories block” for this. Maybe like this?:

image

Something about those serifs feel out of place when mixed in-between comments for me.

nielslange:
@kjellr Thanks for your feedback. I’ll wait for the outcome of your discussion regarding @allancole’s design proposal.

joyously:
Trackbacks separate?

nielslange:
> Trackbacks separate?

@joyously What do you mean by `Trackbacks separate`?

joyously:
I just wondered if the trackbacks are styled the same, and if they are printed separately.
And now that I look again, aren't links supposed to look like links?

kjellr:
Let's just keep this PR simple, and use the styles mentioned here:

https://github.com/WordPress/twentynineteen/pull/336#pullrequestreview-168833710

... along with the edits described in that comment. 

We can discuss trackbacks in a separate comment if we want to print them here.

kjellr:
Made some small adjustments: 

- Removed the singular pingbacks label due to potential translation issues.
- Updated the font styles to more closely match @allancole's comp. But kept the links blue, and bumped "Pingback" label onto its own line for simplicity's sake.

screen shot 2018-11-01 at 1 47 32 pm

Allan, mind giving this a review?

PR #337, opened 2018-10-26
[WIP] Implement designs for mobile nav menu

### LINKED ISSUE #280: Implement mobile menu styles and functionality
#192 added base support for submenu items.  also proposed an approach for treating the submenu on smaller screens. We still need to build in full-screen styles for the submenus on small screens, and also the (likely JS-powered) close and back buttons necessary on the mobile equivalents of these menus. 

(Related to #268, which is to implement the priority+ navigation in JS)

_Comps:_

47155399-a049c180-d2b2-11e8-97f5-f1d1583fd26c

_GIF of the concept in action:_

47155425-afc90a80-d2b2-11e8-8b51-521ec065f9bf

westonruter:
This is a work in progress. 

To start with, it adds the design's required nav menu item to move the user back from an opened submenu to the parent item. 

Comps ror reference:

47155399-a049c180-d2b2-11e8-97f5-f1d1583fd26c

So to begin with it copies the parent nav menu item into the submenu as the first item, and then prepends that nav menu item with a left arrow:

image

This parent nav menu item only is displayed when on a mobile (tablet) breakpoint.

More styling and functionality is needed.

- [ ] Make sure appropriate arrows respect RTL.
- [ ] On mobile breakpoint, show submenu items as sidebars animating from the left with a `translate()` `transform`.
- [ ] Implement X button to close the entire tree of sub nav menus without having to navigate up each level.

Fixes #280.


westonruter:
Here's what I'm thinking in terms of how mobile users can navigate from the top level to submenu items and their nested submenu items. We can add `id` attributes to the link for each parent item, and then in the submenu the link in the submenu (e.g. "< Submenu" and "< Dogs" in the comps above) can have an anchor link that moves the keyboard focus to the "Submenu" and "Dogs" links respectively among the parent items:

image

I'm thinking then we could potentially use `:focus-within` and `:target` selectors to handle displaying the appropriate nav submenus.

westonruter:
And to facilitate opening the submenu to begin with, the mobile menu submenu expansion button could link back to the submenu item, which then itself links back out to the parent item, like so:

image

westonruter:
Alright, I think I've gotten about as far as I'll be able to get for styling (not my forté). Now when clicking a submenu expander on mobile you'll see a fixed sidebar appear on the right:

image

More work is needed to figure out how to get the nested submenu to overlay its entire parent submenu when it is expanded. 

Additionally, you can see that it _kinda_ works without without JavaScript but it feels very janky. The use of the anchor links to hack around with `:focus` results in the page scrolling jump around which is not acceptable. 

I suggest adding a minimal amount of script to handle link clicks which then `preventDefault` and then toggle the appropriate class on the elements and move keyboard focus where it needs to go.

If JS is required for this experience as well, then perhaps this sidebar menu experience should disabled if `html.no-js`.

kjellr:
👍 Thanks, @westonruter! I'll jump in tonight and polish the styling up a bit.

kjellr:
Made some good progress on the styling tonight. Here's what this looks like right now:

**Desktop:**

- Fixed up some spacing
- Swapped the tiny dropdown icon for a chevron (It think this is a little more simple and sophisticated)
- Added a slight fade-in animation to the submenus (It is slowed down a bit below so the GIF framerate picks it up)

desktop-menu


**Mobile:**

- Made the submenus cover the whole screen
- Replaced the arrows with chevrons too
- Added slide-in animations (I can probably wrangle some slide-out animations for when you go up a level in the nav, too)

mobile-menu

The general interaction is still a little bit messy, but this definitely seems like we're getting closer. 👍

Issue #29, opened 2018-10-16
Support for non-latin font fallbacks


waviaei:
Looking at the [_fonts.scss](https://github.com/WordPress/twentynineteen/blob/master/sass/variables-site/_fonts.scss), most of the specified fonts -- as I am aware -- only support Latin alphabets.  They do not support or only very limited support for non-latin alphabets.

In addition to that, serif fonts are specified as the primary body font. 

`$font__body: "Hoefler Text", "Baskerville Old Face", Garamond, "Times New Roman", serif;`

In Japanese, non-serif fonts are more commonly used on the web. If serif fonts are used, we need to consider the choice of them very carefully, else readability and aesthetics become absolutely terrible.

Are there any similar cases in any other languages? I am not sure.

The same issue regarding better support for non-latin font fallbacks was raised and applied in the development of twentyseventeen https://github.com/WordPress/twentyseventeen/issues/59

I would like to raise this issue for same matter to be considered in Twentynineteen too. 

---

_19-Oct Edit by designsimply_

List of alphabets supported in `twentyseventeen` that we also want to try support:

1. Cyrillic ( https://github.com/WordPress/twentyseventeen/pull/364 )
1. Greek ( https://github.com/WordPress/twentyseventeen/pull/366 )
1. Hebrew ( https://github.com/WordPress/twentyseventeen/pull/294 )
1. Arabic ( https://github.com/WordPress/twentyseventeen/pull/398 )
1. Vietnamese ( See [twentyseventeen/59 (comment)](https://github.com/WordPress/twentyseventeen/issues/59#issuecomment-254065128) and [twentyseventeen/59 (comment)](https://github.com/WordPress/twentyseventeen/issues/59#issuecomment-260561023) )
1. Thai ( https://github.com/WordPress/twentyseventeen/pull/93 )
1. Japanese ( https://github.com/WordPress/twentyseventeen/pull/155 )
1. Korean ( https://github.com/WordPress/twentyseventeen/pull/286 )
1. Chinese ( https://github.com/WordPress/twentyseventeen/pull/359 )
1. Devanagari ( https://github.com/WordPress/twentyseventeen/pull/403 )
1. Gujarati ( https://github.com/WordPress/twentyseventeen/pull/404 )

Checklist for `twentynineteen` PRs:

- [ ] Cyrillic (~#343~ #383)
- [ ] Greek (~#344~ #383)
- [ ] Hebrew (~#345~ #383)
[...]

PR #341, opened 2018-10-26
Add Japanese language support

nielslange:
Add Japanese (js: 日本語) language support due to #29.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -145,6 +145,12 @@ function twentynineteen_scripts() {
 	if ( is_singular() && comments_open() && get_option( 'thread_comments' ) ) {
 		wp_enqueue_script( 'comment-reply' );
 	}
+
+	switch ( get_locale() ) {
+		case 'ja':
+			wp_enqueue_style( 'twentynineteen-japanese', get_template_directory_uri() . '/language/japanese.css' );
+			break;
+	}
 }
 add_action( 'wp_enqueue_scripts', 'twentynineteen_scripts' );
 
--- a/language/japanese.css
+++ b/language/japanese.css
@@ -1,0 +1,77 @@
+body,
+button,
+input,
+select,
+optgroup,
+textarea,
+.page-title,
+.entry-content .wp-block-verse {
+  font-family: "Helvetica Neue", helvetica, arial, sans-serif;
+}
+
+.main-navigation,
+.page-description,
+h2.author-title,
+.author-description .author-link,
+.not-found .page-title,
+.error-404 .page-title,
+.post-navigation .post-title,
+.pagination .nav-links,
+.comments-title,
+.comment-author .fn,
+.no-comments,
+.site-title,
+h1, h2, h3, h4, h5, h6,
+.site-info,
+.page-description,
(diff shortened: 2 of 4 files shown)


mirucon:
Why would you use "Hiragino Kaku Gothic Pro" for this? To talk about some details about Japanese font in macOS, there are 3 kinds of the Hiragino Kaku Gothic font (macOS default Japanese font) bundled in macOS Mojave, "Hiragino Kaku Gothic Pro", "Hiragino Kaku Gothic ProN", and "Hiragino Sans". The "ProN" font has created based on JIS2004, the newer standard of Japanese character compared to the base of "Pro". And "ProN" is the default font from OS X 10.5 and iOS (forgot what version, but probably just as old as OS X 10.5) to OS X Yosemite (10.10) and iOS 8. 

And the "Hiragino Sans" was introduced in macOS El Capitan (10.11), the same time of the San Francisco font was introduced, and it's been the default macOS Japanese font since then. And the big advantage of Hiragino Sans is it covers weight 100 to 900 although Pro/ProN only cover 300 and 600 weight (iOS does not have all of them though, but it has more than Pro/ProN). So I would not recommend using either "Pro" or "ProN". They are not 2019's.

However just specifying "Hiragino Sans" in font-family isn't really the answer in my opinion. I personally do not like the idea of using Hiragino fonts alphabets, because it's not as good as other alphabet fonts bundled in macOS. What I suggest is to use Hiragino Sans font with San Francisco (which can be specified via -apple-system and BlinkMacSystemFont). It's the pair of San Francisco and Hiragino Sans is just same as macOS system default so people should have gotten used to it.

Therefore this is what I suggest using for Japanese: 

```
-apple-system, BlinkMacSystemFont, "Helvetica Neue", "Hiragino Sans", Meiryo, sans-serif;
```

And use this for the theme body including input and textareas.

The opinion about the Hiragino alphabets could be very personal, though, I believe the idea of using "Hiragino Sans" is a correct thing in 2019.

nielslange:
@mirucon Thanks for your feedback. I used that particular font based on a discussion that took place in https://github.com/WordPress/twentyseventeen/pull/155. I‘m happy to adjust the font according to your suggestion. Do you know if Windows systems support "Hiragino Sans" as well? And wouldn’t it be better to use the following order:

```
-apple-system, BlinkMacSystemFont, "Hiragino Sans", Meiryo, "Helvetica Neue", sans-serif;
```

mirucon:
Unfortuately the Hiragino Sans is not supported in Windows. However "Meiryo" is the one bundled in Windows Vista or later. Some people might argue with whether we should use Yu Gothic (which it's bundled in Windows 8.1+) though, I'm against using this font for websites since it looks too thin on some browsers, especially screens with low resolution.

Since the meiryo fonts was designed as the extended version of Verdana, the alphabets of this font is not bad and I don't really think we should use separate font for alphabets in Meiryo. So what you suggested looks good to me :)

nielslange:
@mirucom Thanks for letting me know. I’ll make that change later today.

nielslange:
Closed in favour of #383

PR #344, opened 2018-10-26
Add Greek language support

nielslange:
Add Greek (el: Ελληνικά) language support due to #29.

### DIFF
--- a/language/greek.css
+++ b/language/greek.css
@@ -1,0 +1,77 @@
+body,
+button,
+input,
+select,
+optgroup,
+textarea,
+.page-title,
+.entry-content .wp-block-verse {
+  font-family: "Helvetica Neue", Helvetica, "Segoe UI", Arial, sans_serif;
+}
+
+.main-navigation,
+.page-description,
+h2.author-title,
+.author-description .author-link,
+.not-found .page-title,
+.error-404 .page-title,
+.post-navigation .post-title,
+.pagination .nav-links,
+.comments-title,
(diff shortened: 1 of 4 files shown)


language/greek.css
```diff
@@ -0,0 +1,77 @@
...
+.entry-content .wp-block-cover h2,
+.entry-content .wp-block-audio figcaption,
+.entry-content .wp-block-video figcaption,
+.entry-content .wp-block-image figcaption,
+.entry-content .wp-block-gallery .blocks-gallery-image figcaption,
+.entry-content .wp-block-gallery .blocks-gallery-item figcaption,
+.entry-content .wp-block-file,
+.entry-content .wp-block-file .wp-block-file__button,
+.entry-content .wp-block-latest-comments .wp-block-latest-comments__comment-meta,
+.gallery-caption {
```
grappler: Is there not a better way then to target nearly every class?

nielslange: @grappler I copied all affected statements from `style.css`. If I shorten the statements in `greek.css` they get overwritten by `style.css`. One way to avoid that the statements get overwritten would be to force the statements using `!important`.

I also thought about outsourcing the localised fonts into `/languge/_font-greek.scss`, copying the contents from `style.scss` into `language/greek.scss` and overwriting the font settings. However, this will create exact almost clones of `style.css` itself, which is also not very smooth. 

How would you solve the particular issue above?

nielslange:
Closed in favour of #383

PR #359, opened 2018-10-26
Remove underline from file download button text

### LINKED ISSUE #358: File download button has underline under text
See screenshot below

file-download-button-underline

ianbelanger79:
Fixes #358

### DIFF
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -547,6 +547,7 @@
 			line-height: $font__line-height-heading;
 			font-weight: bold;
 			padding: ($size__spacing-unit * .75) $size__spacing-unit;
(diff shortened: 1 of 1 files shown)


ianbelanger79:
@kjellr I combined this PR with #360, so they should be good to go

sass/blocks/_blocks.scss
```diff
@@ -135,6 +135,7 @@
 			outline: none;
 			color: white;
 			outline: none;
+			text-decoration: none;
```
kjellr: This button underline was actually taken care of separately via #363 earlier today (Sorry about that — I hadn't realized your PR was tackling all buttons globally). This will add a duplicate rule to the file, so it would need to be removed.

PR #360, opened 2018-10-26
Remove underline from file download button in style.css

### LINKED ISSUE #358: File download button has underline under text

ianbelanger79:
Fixes #358 

WP.org username @ianbelanger


kjellr:
Hi @ianbelanger79 — would you mind tying this change directly into #359? That way we won't have to review and merge two separate PRs. 

Thanks!

Issue #342, opened 2018-10-26
When user Avatar is there entry meta is breaking down on Single Post Template


monika-12:
Entry meta is breaking down in 2 lines when Avatar is there.

![screenshot-wp local-2018 10 26-12-57-31](https://user-images.githubusercontent.com/17018981/47551054-c01a5e80-d91e-11e8-9f1d-f21dc73bf2de.png)

kjellr:
Thanks. I can reproduce in Safari 12.0. For some reason, this does not occur when I'm logged in, only when I'm logged out.

<img width="1440" alt="screen shot 2018-10-26 at 8 29 47 am" src="https://user-images.githubusercontent.com/1202812/47566315-5c316f00-d8f9-11e8-98f1-16834444955e.png">

PR #367, opened 2018-10-26
 Fix comment avatars in singular featured-image-headers

allancole:
Fixes: #342 

Also addresses a few other issues: 
• Unifies entry-header markup in all singular views with & without featured images
• Cleans up discussion avatar displays in header and in comments section
• Fixes text wrapping in .entry-meta to keep icons and link-text together at all times


kjellr:
@allancole: One quick issue I noticed: the avatars can overlap the title sometimes: 

screen shot 2018-10-26 at 8 17 03 pm

If there are comment avatars, we should probably have a slightly narrower max-width on the title to avoid this.

PR #380, opened 2018-10-28
Fixed media alignment

shahjehanali1:
Fixed center alignment issue. If image has caption. 

sample page onepagepallax - google chrome 2018-10-29 00 45 50


kjellr:
Shortly after merging, I realized this style was actually only needed on our `desktop` breakpoints, not our `tablet` ones. Otherwise it  interfered with already-working styles at the breakpoints in between tablet and desktop. Pushed a quick update via https://github.com/WordPress/twentynineteen/commit/fdf944f68abc984e87c9adbe5fc1befea1b32e61. 👍

PR #397, opened 2018-10-30
Replace word-break property with word-wrap

### LINKED ISSUE #395: word-break is messing up the text
Someone added `word-break: all` in a recent PR. It is messing up the content.
I think the margins are too large for the images, but still, the words should not be broken like that.

word-break

kjellr:
Fixes #395

Word-break is too aggressive for what we want to do here, which is just prevent the overflow of long unbroken strings or URLs.

I attached this just to paragraphs within entry-content for now, to be conservative about its usage.

Before:
screen shot 2018-10-29 at 9 41 47 pm

After:
screen shot 2018-10-29 at 9 42 01 pm

### DIFF
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -5,7 +5,6 @@
 
 	margin: 32px $size__spacing-unit;
 	max-width: calc(100vw - (2 * #{ $size__spacing-unit }));
-	word-break: break-all;
 
 	@include media(tablet) {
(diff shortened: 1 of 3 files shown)


allancole:
Looks good. Merging…

kjellr:
Also noting that I _could've_ used the newer name, `overflow-wrap` in here, but IE11 and Edge require `word-wrap`. Since that `word-wrap` works in all browsers, so it seemed to be the simplest option for now.

Issue #50, opened 2018-10-16
Do a full pass on responsive images


mor10:
With full-width images being a thing in this theme it is essential we do a full pass on responsive images to ensure the browser has sufficient info (ie the correct `sizes` attributes) to load the image size it needs for the current display.  This requires testing. TBC.

kjellr:
In general, `sizes` attributes for full-width images should be handled in Gutenberg. As long as this is covered there, we should be set. 

Wide width featured images are another story though- so I'll mark this as a `task` to follow up on.

mor10:
I'll do some testing to see if any issues arise. Earlier versions of Gutenberg were not handling this well, but I know work has been done so hopefully it's a non-issue.

LittleBigThing:
Related to this issue. We are using a background image as a featured image for single posts and pages with `background-size: cover`. This means that the same image is loaded for each screen size and we are not taking advantage of responsive images as available in core for the `<img>´ tag.

As an alternative, we could output the collection of images in `<img>` and use `object-fit: cover` to style it. `Object-fit` has a quiet [good browser support](https://caniuse.com/#search=object-fit), except for IE.
It was also used for Twenty Seventeen, so we have an example. 👍

I think that this is still doable and that it could make a difference for mobile/tablet usage. Although I'd need to test it first regarding the actual gains as that depends on the available image sizes and how much an images has to 'be fitted' to have it cover the whole screen.

Also related to #181.
This can also be related to supporting header media (like images and videos) if requested.

joyously:
Be sure to consider #208.

LittleBigThing:
Ok, thanks. The issue of having small featured images will probably remain using the above option.

kjellr:
@LittleBigThing A refactor to use images that way would be fantastic. I'm not sure @allancole or I will have a chance to work on that before RC1, but I'd definitely encourage any PRs.

mor10:
I'll take a closer look on Monday. We can leverage CSS Grid to make the featured image responsive and accessible with a background image fallback.

mor10:
I've done some preliminary testing and I think work needs to be done here. At present, responsive images are not handled correctly resulting in a significant performance hit. How to solve this problem depends in part on the result of [Gutenberg issue 6177](https://github.com/WordPress/gutenberg/issues/6177). Tagging @azaozz for input.

To replicate my procedure: 
1. Add a single 1600px wide image to a post using the "normal" displayed width (alignment doesn't matter).
[...]

PR #411, opened 2018-10-30
Add `sizes` attribute for responsive images markup

mor10:
Addresses #50. 

Prototype. Currently only works properly for "regular" width images. 

Notes: 
- `$GLOBALS['block_width']` section needs work as the theme uses fluid widths for all images on wider screens. Current values are effectively random.
- To provide correct `sizes` attributes for `alignwide` and `alignfull` images we need a way of knowing, from within the `wp_calculate_image_sizes` filter, what the width display setting is for the current image. At present it is only defined in a class on the parent `<figure>` element which to my knowledge is inaccessible from the filter.

Works as prototype for testing of [Gutenberg issue 6177](https://github.com/WordPress/gutenberg/issues/6177). 

cc @azaozz

Todo:

- [ ] `sizes` attributes for `alignwide` and `alignfull`
- [x] `sizes` attribute for featured images

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -152,7 +152,7 @@ function twentynineteen_content_width() {
 		'default' => 723,
 		'wide'    => 800,
 		'full'    => 1024,
-		);
+	);
 }
 add_action( 'after_setup_theme', 'twentynineteen_content_width', 0 );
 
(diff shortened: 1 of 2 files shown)


mor10:
`sizes` attribute for featured images added and merged in #493.

mor10:
Closing in favor of #629

PR #415, opened 2018-10-30
Fix/gravatar alternative text

### LINKED ISSUE #276: Consider to improve the comment author info accessibility
When a comment author has filled in the website info, Twenty Nineteen makes the avatar a link to the comment author website. However, the only content of the link is the avatar image with an empty alt attribute `alt=""`. It's basically an empty link and there's no meaningful information assistive technologies (and search engines crawlers) can use for the link other then the link URL.

Screen readers will try to announce _something_ and the only available content to announce is the link URL:

screen shot 2018-10-24 at 17 28 16

An option could be using a proper text for the alt attribute, something like: `{comment author name} website`.

Even better, as there are two adjacent links that point to the same destination (the avatar and the visible, bold, author name):

screen shot 2018-10-24 at 17 50 39

I'd propose to combine the two links, just moving the image within the bold link, meeting this simple WCAG technique: https://www.w3.org/TR/WCAG20-TECHS/H2.html

A second issue in this markup is about the elements order:
1. comment author image + name
2. visually hidden text "says:"
3. comment meta data: date + time
4. finally, the comment content

Software read content in a linearized way. With the current order, a screen reader would read out, for example:

> link, John Doe  says: link, October 24th, 2018 at 6:03 pm ...

As the visual order of the elements must always match the DOM order, and I guess visual changes are not desirable at this point, I'd suggest to just remove the visually hidden text `says:` which adds little or no benefit to users.

anevins12:
Resolves  #276


anevins12:
@allancole  Are you sure that's not already on master?

PR #422, opened 2018-10-31
🐛Fix fatal error in php5.5 or less

Artisan-Asad:
Fatal error in php version 5.5 or less in site-branding.php L16
https://monosnap.com/file/gAKJJaSCPzYHncz1TmJ78uBayGQZge

and same situation was encounterd in footer.php L21
https://monosnap.com/file/gAKJJaSCPzYHncz1TmJ78uBayGQZge

Problem: Only variables can be passed to empty() prior to PHP 5.5


nielslange:
@Artisan-Asad Thanks for your contribution. 

@allancole & @kjellr According to https://wordpress.org/about/stats/ 19.5% of all WP installs are running on PHP < 5.5. Based on that high percentage and the fact that the files `site-branding.php` and `footer.php` are affected, I would merge this PR (_despite the fact that I think we should not support PHP versions that are [no longer officially supported](http://php.net/supported-versions.php)_).

ocean90:
Just [like WordPress itself](https://wordpress.org/about/requirements/), the default theme must support PHP 5.2.4+.

Artisan-Asad:
👍 Your welcome @kjellr  and @nielslange looking forward to see you in Word Camp Islamabad 😉

PR #434, opened 2018-10-31
Fix placement of translators: comments

Ismail-elkorchi:
For translators comments to be recognized, they must be on the line directly before the gettext call.

### DIFF
--- a/classes/class-twentynineteen-walker-comment.php
+++ b/classes/class-twentynineteen-walker-comment.php
@@ -65,10 +65,10 @@ class TwentyNineteen_Walker_Comment extends Walker_Comment {
 						<a href="<?php echo esc_url( get_comment_link( $comment, $args ) ); ?>">
 							<?php
 								/* translators: 1: comment date, 2: comment time */
-								$comment_timestamp = __( '%1$s at %2$s', 'twentynineteen' );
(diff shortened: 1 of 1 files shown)


Ismail-elkorchi:
Another solution is to store the printf ( since it's called twice ) in a variable and echo it.

The code will look like something like this : 
```
<?php
	/* translators: 1: comment date, 2: comment time */
	$comment_timestamp = printf( __( '%1$s at %2$s', 'twentynineteen' ), get_comment_date( '', $comment ), get_comment_time() );
?>
<time datetime="<?php comment_time( 'c' ); ?>" title="<?php echo $comment_timestamp; ?>">
	<?php echo $comment_timestamp; ?>
</time>
```

joyously:
It would have to change from printf to sprintf.

Ismail-elkorchi:
> It would have to change from printf to sprintf.

You're right! I changed my comment to reflect that.

Ismail-elkorchi:
@kjellr Can you please review this PR, and tell me which approach is the best. It's the one committed in the PR, or the one suggested in the comment above ? So I can update this PR.

allancole:
> Another solution is to change the `printf` call to `sprintf`, store it in `$comment_timestamp` ( since it's called twice ) and echo it.

Yup, I think this the right way to go @Ismail-elkorchi . Make that change and we’ll go ahead and merge this in 👍

Ismail-elkorchi:
Thank you @allancole for your review. I made the change requested, it should be good to merge 👍

Issue #398, opened 2018-10-30
Menu subitems can extend offscreen on desktop


kjellr:
Menu items can extend beyond the right side of the viewport. This happens especially frequently with submenu items.

![menu](https://user-images.githubusercontent.com/1202812/47690972-a7cc6d00-dbc6-11e8-82a7-a68bee378fd2.gif)

joyously:
What happened to the menu showing only the items that fit? (I know that doesn't apply to submenu.) 
If you make your window smaller, the menu is wrapping around instead of having a `...` like it was proposed.

kjellr:
> What happened to the menu showing only the items that fit?

That's still the plan, and is  an open PR (#268). It needs to be reworked a little so that it can be implemented alongside the menu updates that were pushed today.

shahjehanali1:
@kjellr I've fixed this issue https://www.useloom.com/share/e368a9591e3644cea0dd82df45a7bd17

 but not b able to push code me in 'add/priority-menu-two'

shahjehanali1:
https://gist.github.com/shahjehanali1/c8251b55b4fabc97edcf5f698958034b

kjellr:
This should be generally taken care of via the updates in #485:

- Submenu items within dropdown menus are now exposed by default. We no longer use submenus that expand off to the right. 
- The Priority+ navigation button eats up top-level menu items that would otherwise cause overflow.

Those updates should handle the vast majority of potential problems here, so I'm going to close this issue for now. Please submit new issues if any other additional menu overflow problems are discovered.

PR #437, opened 2018-10-31
Improve priority+ menu with, keyboard nav and touch device support

allancole:
Combines #337 and #313

• Adds priority+ UI
• Adds support for keyboard navigation 
• Also supports `:hover` states on desktop screens
• Adds supports for touch devices 
• Fixes an issue where priority+ would require resizing to instantiate

**Edit**
Adding this checklist borrowed from an earlier comment here: https://github.com/WordPress/twentynineteen/pull/313#issuecomment-434425130

- [x] In my testing, the priority+ menu didn't consistently kick in on every new page load. Sometimes I had to resize the window a little bit for it to activate. 
- [ ] For now, I exposed and indented all submenus in this iteration. I'm thinking we'll either need to replace that with the nested solution we have from #337 ([GIF here](https://github.com/WordPress/twentynineteen/pull/337#issuecomment-433585675)), or we'll need to adapt this method for all menu items. (It's simple and it saves users a tap or two).
- [x] On mobile, you can currently scroll and see the normal page under the menu. 
- [ ] On mobile, we need to inject a close button.
- [x] Our current desktop menus work on hover, but this requires a click. 
- [x] On desktop, the only way to close this is to click the ellipsis button again. We should have it close whenever someone hovers outside of the menu (Or clicks, depending on whether or not we use a click or a hover to trigger the menu, as per the above bullet point). As it stands now, you can activate both a submenu _and_ the priority+ menu, which leads to [some messiness](https://cloudup.com/cv-Oi-JbwCr). 
- [x] On desktop, the menu is just anchored to the right of the header area. This is usually okay, but it'd be better if this was more closely aligned with the button itself. 
- [ ] This should be tested more. 😄 

Does not include a solution for this bug #398 

**Screenshot**

2019 tabbed menu 2


kjellr:
In addition to the notes above, I noticed a few things testing just now:

- The Priority+ menu is visible all the time, even when it's empty. 
- Priority+ menu items with a submenu are [showing a down-arrow on large screens](http://cld.wthms.co/WlquR4), but they expand to the right.
- On Safari (on a laptop), the menu isn't usable at all below 768px. Clicks don't register on the Priority+ button.

allancole:
Closing this in favor of #485

PR #447, opened 2018-10-31
#429: Amended alignment to dot separator

### LINKED ISSUE #429: Dot separator not centered in content, front and backend
When a Dot Separator is added, it is not centered within the content on the front or backend. See screenshots below.

Frontend
dot-separator-not-centered

Backend
dot-separator-not-centered-backend

anevins12:
Resolves #429

### DIFF
--- a/sass/mixins/_mixins-master.scss
+++ b/sass/mixins/_mixins-master.scss
@@ -42,7 +42,7 @@
 		mix-blend-mode: multiply;
 		opacity: .8;
 		z-index: 3;
-		
+
 		/* Browsers supporting mix-blend-mode don't need opacity < 1 */
 		@supports (mix-blend-mode: multiply) {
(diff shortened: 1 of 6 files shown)


sass/mixins/_mixins-master.scss
```diff
@@ -87,4 +87,17 @@
 		0 0 1px 0 rgba( $color__text-light, 0.7 );  	// Antialiasing, outer edge.
 }
 
+/* Calculates maximum width for post content */
```
kjellr: If you have some time, it'd be great to take a look around the codebase to see where else this mixin might be useful. We could probably reuse this more frequently if it accepted variables. For example: `postContentMaxWidth( $mobile: 2, $tablet: 8, $desktop: 6)`. That's all a separate PR though. 🙂

PR #448, opened 2018-11-01
Add composer and update phpcs.xml.dist

dingo-d:
To run the phpcs check, run 
```bash
composer check-cs -- --ignore=/vendor/ .
```

to fix issues automatically run

```bash
composer fix-cs -- --ignore=/vendor/ .
```


composer.json
```diff
@@ -0,0 +1,31 @@
...
+	],
+  "require": {
+      "wptrt/wpthemereview": "^0.1.0"
+  },
+  "require-dev": {
+      "dealerdirect/phpcodesniffer-composer-installer": "^0.5.0"
+  },
+  "scripts": {
+		"phpcs-i": "@php ./vendor/squizlabs/php_codesniffer/bin/phpcs -i",
+		"check-cs": "@php ./vendor/squizlabs/php_codesniffer/bin/phpcs",
```
ntwb: Can this use `lint` as the script name please, this matches cores usage.

https://core.trac.wordpress.org/browser/trunk/composer.json#L17

composer.json
```diff
@@ -0,0 +1,31 @@
...
+		"fix-cs": "@php ./vendor/squizlabs/php_codesniffer/bin/phpcbf"
```
ntwb: Can this use `format` as the script name please, this matches cores usage.

https://core.trac.wordpress.org/browser/trunk/composer.json#L18

composer.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "wordpress/twentynineteen",
+  "description": "A new Gutenberg-ready theme.",
```
ntwb: This description differs from the description in `package.json` which is currently `Default WP Theme`, taking the description from the themes `style.css` seems best, `Twenty Nineteen`

https://github.com/WordPress/twentynineteen/blob/master/style.css#L3

composer.json
```diff
@@ -0,0 +1,31 @@
...
+  "type": "package",
+  "license": "GPL v2 or later",
+  "keywords": [
+		"WordPress",
+		"Themes"
+	],
+  "authors": [
+		{
+			"name": "Contributors",
```
ntwb: The authors name should also be taken from the `style.css` file, `the WordPress team`

https://github.com/WordPress/twentynineteen/blob/master/style.css#L5

composer.json
```diff
@@ -0,0 +1,31 @@
...
+			"homepage": "https://github.com/WordPress/twentynineteen/graphs/contributors"
```
ntwb: Also, per `style.css`, the home page should be `https://github.com/WordPress/twentynineteen`

https://github.com/WordPress/twentynineteen/blob/master/style.css#L4

ntwb:
Nice @dingo-d, adding the `vendor` folder as an _exclude_ entry in the `phpcs.xml.dist` file would also be preferable rather than a cli flag please

composer.json
dingo-d: So should `style.css` be updated to match `package.json` description (`Default WP Theme`) or vice versa - should I update `package.json` to say `A new Gutenberg-ready theme.`?

Twentyseventeen had a totally different description

> Twenty Seventeen brings your site to life with immersive featured images and subtle animations. With a focus on business sites, it features multiple sections on the front page as well as widgets, navigation and social menus, a logo, and more. Personalize its asymmetrical grid with a custom color scheme and showcase your multimedia content with post formats. Our default theme for 2017 works great in many languages, for any abilities, and on any device.

I think that this should be changed in both `package.json`, `composer.json` and `style.(s)css` once the theme is ready to be published 🙂

dingo-d:
Thanks for the comments @ntwb, I've fixed all except the description one as I think a better description should be used and made uniform across three files.

Also, a question about travis failing: how come you are installing WPCS using `wget` and `curl` instead `composer require`? Like here: https://github.com/WPTRT/WPThemeReview/blob/develop/.travis.yml#L54

composer.json
ntwb: Not sure, both the `composer.json` and `package.json` files probably don't require a long description.

Thoughts @allancole @kjellr ?

ntwb:
> Also, a question about travis failing: how come you are installing WPCS using wget and curl instead composer require? Like here: WPTRT/WPThemeReview:.travis.yml@develop#L54

It was a quick copy and paste from another repo, iterating on this is fine, so have at it in another PR if you'd like, please, and thank you 😏

composer.json
kjellr: @dingo-d Let's just use the description that we've been putting in the core commit notes:  

> Let Gutenberg shine with this simple, fast, and powerful theme.

This will be revised and expanded for the actual launch, but that'll work for now.

allancole:
@dingo-d could you give a more detailed description of what this does? It isn’t tied to a ticket and i’m not exactly sure how necessary it is. Seems like it would suppress some of the Travis.ci failures, but its not likely that this would make it into `core`.

dingo-d:
Hi @allancole, this adds a `composer.json`, and modifies the `phpcs.xml.dist` so that the `WPThemeReview` ruleset is used, which is the ruleset used for sniffing themes for theme review at wordpress.org.
Also it provides an easy way to check for coding standards using composer scripts.

Basically, a development help for the theme. Plus since it's an official theme, this can be used as an official way to introduce usage of coding standards when developing themes 🙂You can read more about the ruleset here: https://github.com/WPTRT/WPThemeReview

ntwb:
> Seems like it would suppress some of the Travis.ci failures

This PR in its current state will not change the currently reported errors by Travis CI. 

For that to occur `travis.yml` should also be updated to use this implementation.

This could be added to this PR, or added in a follow up PR.

> but its not likely that this would make it into core.

It actually would not matter is the `composer.json` and `composer.lock` were committed to core, they are "build tool files" extremely similar to the existing `package.json` and `package-lock.json` files [already commited](https://core.trac.wordpress.org/browser/branches/5.0/src/wp-content/themes/twentynineteen)

Putting my [build tools component maintainer](https://make.wordpress.org/core/components/) hat on @allancole it would be good if you could in fact commit these files on the next update commit to core SVN please.

At some stage we need to update the coding standards of all the core themes, and by also adding the "Theme Review Team" PHPCS specifics we can get the ball rolling with the Twenty Nineteen theme and update the remaining themes as part of WP 5.1

allancole:
@ntwb Happy to merge this in but I’m not very familiar with kind of tooling. 

> This PR in its current state will not change the currently reported errors by Travis CI.
> For that to occur travis.yml should also be updated to use this implementation.
> This could be added to this PR, or added in a follow up PR.

If it doesn’t make a huge difference and since all of this seems to be connected, I’d prefer to have the changes in this one PR, so we can get it to pass those current Travis.ci errors. Could you or @dingo-d take care of that? Then we can get this merged :-)

ntwb:
Cool, @dingo-d, can you update the `.travis.yml` file to use the changes in this PR please?

Basically all it needs to do is remove the old PHPCS bits and pieces and run the `lint` Composer script, I think.

dingo-d:
Will do that asap 🙂

dingo-d:
Ok, just a question. It seems that phpcs is only ran on PHP 5.2, but we want to run it on all PHP versions or not?

ntwb:
Composer requires > PHP 5.3, so it needs to switch away from that 5.2 job anyways.

Does WPTRT PHPCS include the PHPCompatibility sniffs?

Down the road I'm sure we'll want to run the PHPCompatibility sniffs, so running PHPCS on all PHP versions > 5.3 sounds good to me

dingo-d:
WPThemeReview has compatibility sniffs included. I'll need to rewrite the travis to use composer on all versions except 5.2, and for 5.2 do a manual install, and check phpcs without composer.

.travis.yml
```diff
@@ -5,108 +5,139 @@
...
 
 # Declare project language and PHP versions to test against.
-# @link http://about.travis-ci.org/docs/user/languages/php/
+# @link https://about.travis-ci.org/docs/user/languages/php/
 language: php
 
 # Newer versions like trusty don't have PHP 5.2 or 5.3
 # https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming
-dist: precise
+dist: trusty
```
ntwb: PHP 5.2 and 5.3 are not available on Trusty, this is why it has remained Precise

dingo-d: I have added them under:
```yml
    - php: "5.3"
      env: PHPCS_BRANCH="2.9.0" SHORT_OPEN_TAGS=true PHAR_LINT=1
      dist: precise
    # Current minimum $required_php_version for WordPress: 5.2.4
    - php: "5.2"
      env: PHPCS_BRANCH="2.9.0" SHORT_OPEN_TAGS=true PHAR_LINT=1
      dist: precise
```
So this should be ok. The travis fails on phpcs `wget` tho.

ntwb: Hehe, nice, ignore me 😏

dingo-d: Also I noticed that phpcs wasn't running in the previous builds...

dingo-d:
Ok, so I've fixed the travis, but I had to remove jshint, because it caused errors. I added js linting via phpcs, and the build fails because there are js errors. Can I fix those and submit them in this PR so that the build passes?

ntwb:
Thanks a bunch @dingo-d, considerable effort and work getting all this up and running, appreciated immensely 😄

dingo-d:
Happy to help 🙂 Let me know if something should be modified or updated, I'll gladly help 🙂

Issue #181, opened 2018-10-19
make the theme as lightweight as possible


LittleBigThing:
Lightweight themes can load fast without spending hours optimising them by adding stuff. With WordPress powering over 30% of all websites, a lightweight theme can make a real difference in sustainable web development. With Gutenberg, it will be easier to create richer layouts without using more complicated and often bloated solutions. This default theme could demonstrate best practice on this area. And seems that it is off to a good start.

Here, we could gather some suggestions to keep/make Twenty Nineteen as lightweight as possible without losing essential features and functionality.

To start with, here are some suggestions:
- jQuery is not used for now, let's keep it this way: **check**
- do not use webfonts: **check**
- use progressive enhancement, of course
- use polyfills to account for essential functionality in non-supporting browsers
- ...

ianbelanger79:
- Use just vanilla CSS instead of SASS, as is being discussed in #30

kjellr:
Thanks for the reminder, @LittleBigThing! 

These are absolutely great points, and keeping the theme lightweight is an important goal. As you noted, this was the driving force behind the theme's reliance on system fonts (and the fact that we don't — and have no plans to use — jQuery). 🙂 Making use of progressive enhancement and accounting for similar or equivalent functionality across our [supported browser set](https://make.wordpress.org/core/handbook/best-practices/browser-support/) is also key. 

You can help by: 

- Identifying areas where we're currently _not_ fulfilling the needs you mentioned above. Your issue #121 is a great example of this. 🙌
- Looking for specific areas of improvement in our codebase, and proposing solutions. For example, (I haven't had a chance to review yet, but) #39 and #47 are great examples of PRs aimed at being smarter about the loading of scripts.

Identifying these issues + opportunities at an early stage is essential to making them happen.

LittleBigThing:
Check out the [latest update](https://github.com/WordPress/twentynineteen/pull/450#issuecomment-435398036) for  #450 as an enhancement, related to this issue.

kjellr:
👋 Since this is more of a point of feedback than an addressable issue, I'm going to close it rather than port it over to Trac. 

Thanks for helping us keep the theme lightweight and speedy, @LittleBigThing!

PR #450, opened 2018-11-01
responsive featured image

LittleBigThing:
Related to #50 and to #181.

First take on implementing the featured image as an `` tag to take advantage of the responsive images `srcset` attribute (using `object-fit`). It does **not work right now on IE11**, but I think we could make it work using @supports and some additional CSS (see Twenty Seventeen).

Definitely an enhancement. The benefit would be that smaller images would be loaded on smaller screens instead of the same large image as with a background image.


kjellr:
This is a fantastic PR — thanks @LittleBigThing!

What problem(s) should we be looking out for in IE? I expected it to be obvious, but it's actually looking okay for me...

---

**IE11:**
screen shot 2018-11-01 at 9 48 18 am
screen shot 2018-11-01 at 9 51 22 am

---

**Edge**
screen shot 2018-11-01 at 9 46 42 am
screen shot 2018-11-01 at 9 52 22 am

---

I also tested in FF, Safari, and Chrome, and on iOS. It's looking great in all those places. 

In any case, if we can get any IE bugs sorted out, I'd love to get this in for RC1.

LittleBigThing:
Try to resize the browser in IE, at some points it is stretching the image. Also, it behaves differently in IE using a portrait image.

I’ll look at the implementation in Twenty Seventeen in more detail when I have some time tomorrow...

LittleBigThing:
I just updated the PR with a fix for IE.
It is borrowed from the solution in Twenty Seventeen and I think it works fine now for all major browsers supported by the theme.

All browsers supporting `object-fit` and `@supports` will use `object-fit` now to fit the featured image just a background image would be fitted (`cover`). IE11 is using a kind of manual fitting using `transform: translate(...)`. Anyway, IE11 and lower do not support `srcset` to take advantage of responsive images.

**Benefit**
I tested the benefits just for a page with a featured image in Chrome:
- it is loading the post-thumbnail sized image at _1920px and 1280_px screen width: **176kb**
- it is loading a smaller image (1024 x 683px) for a screen width of _920px_: **72.6kb**
- an even smaller image(768 x 512px) is loaded at a screen width of _768px and below_: **40.7kb.**

We should probably look at the semantics of the html added. The `` is not wrapped in any additional `div` or `figure` element at this moment... Also, the `.wp-post-image` class of the image is targeted, which may not be ideal (?).

kjellr:
Awesome, thanks again @LittleBigThing. I'll give this a review later today. 👍

mor10:
I think the CSS here could be cleaned up significantly by using `grid` instead of `flex` + other complexities, but as it stands the code looks good on first pass.

kjellr:
I tested in Safari, FF, Chrome, Edge, and IE11 (with featured images using variety of sizes/orientations) and it's looking great in every case so far. 👍 

>  We should probably look at the semantics of the html added. The  is not wrapped in any additional div or figure element at this moment... Also, the .wp-post-image class of the image is targeted, which may not be ideal (?).

I'll cc @allancole for thoughts on that, since he may have some thoughts. 

> I think the CSS here could be cleaned up significantly by using grid instead of flex + other complexities, but as it stands the code looks good on first pass.

Agreed. That could be handled separately though, since this PR doesn't really modify the flexbox styles. 

Barring those semantic questions, I think this is likely good to go as a first pass. It's a clear improvement to what we've got. 👍

kjellr:
> Did you try it with a small featured image?

I did test that, and it appears near-identically to the way they appear right now.

allancole:
> We should probably look at the semantics of the html added. The  is not wrapped in any additional div or figure element at this moment... Also, the .wp-post-image class of the image is targeted, which may not be ideal (?).

Yup, I think we should add a `<figure>` element on this as long as it doesn’t break things 👍

LittleBigThing:
Just updated the PR to use `twentynineteen_post_thumbnail()` to output the featured image. This seemed more consistent. This means that `` is now wrapped in `<figure class="post-thumbnail">`.

I checked on Chrome, IE and Edge, haven't noticed any changes.

One drawback is maybe that there is now a double check for `is_singular() && twentynineteen_can_show_post_thumbnail()`.

mrwweb:
> I think we should add a <figure> element on this as long as it doesn’t break things

I'm not sure I agree.

Is this image really content or is it decorative? It feels more decorative to me, but I could see people disagreeing. If it's decorative (which might be why it was implemented as a background image in the first place), then applying `aria-hidden="true"` to the markup might actually be a better solution.

From [MDN's `<figure>` docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figure):

> he HTML <figure> element represents self-contained content, frequently with a caption (<figcaption>), and is typically referenced as a single unit.

> Usually a <figure> is an image, illustration, diagram, code snippet, etc., that is referenced in the main flow of a document, but that can be moved to another part of the document or to an appendix without affecting the main flow.

Issue #187, opened 2018-10-19
Editor styles strange spacing


jeroenrotty:
Making up a first post while twentynineteen is active, placing a space has like double width until you type a letter. Something is buggy probably in the editor styles.
![image 2018-10-19 at 2 57 21 pm](https://user-images.githubusercontent.com/7323966/47219535-601c3900-d3af-11e8-8318-f8e5b41f3ca7.png)

Too clarify, this is on a local dev with only Gutenberg active, fresh install. Doesn't happen on twentyseventeen.

kjellr:
@jeroenrotty — I gave this a try today and was unable to reproduce. 🤔 

![screen shot 2018-10-23 at 8 45 06 am](https://user-images.githubusercontent.com/1202812/47361472-044bfb80-d6a0-11e8-9d5a-f80558d91360.png)

Can you share what browser and operating system you're seeing this with? Thanks!

richsalvucci:
I can replicate with Chrome 69 and Mac OS Sierra 10.12.6. I would type a word, hit the space bar and then it would jump back after I resumed typing. I also tested with Fire Fox 62 and the issue was not happening.

kjellr:
Ah, I can see it in Chrome 69, too:

![space](https://user-images.githubusercontent.com/1202812/47379933-47bb5f80-d6ca-11e8-9764-168175062e23.gif)

This is an odd one. When you insert a space, the editor adds `&nbsp;` to the end of your line. My guess is that for some reason, that non breaking space is styled too be wider than the rest of the spaces in the sentence. We'll need to do some digging to figure this one out.

jeroenrotty:
@kjellr I'm using the latest OSX and latest Chrome (both stable)

jasmussen:
Reported this as an upstream issue: https://github.com/WordPress/gutenberg/issues/11405 

I don't think there's a fix in TwentyNineteen, outside of swapping out the font :(

kjellr:
Closed via #463

PR #463, opened 2018-11-02
Add a font family overwrite for nonbreaking spaces

kjellr:
Attempts to address #187: A weird bug in Chrome where Hoefler text `&nbsp` characters are rendered as wider than a standard space in the editor.

Props to @iseulde 🙌


style-editor.scss
```diff
@@ -12,6 +12,12 @@ when Gutenberg supports styling those variations more intuitively.
 @import "sass/variables-site/colors";
 @import "sass/mixins/mixins-master";
 
+@font-face {
```
iseulde: It's probably wise to add a comment to this hack. :)

jasmussen:
Holy guacamole, this hack is amazing! I'm mobile at the moment but this is an amazing amazing trick and I can't see why it shouldn't work. (Only thing would be to check that the width of this nbsp character set in Georgia is close enough in width to Hoefler that it looks the same)

But wow I just want to give a silent emoji round of applause for this code, it's delicious! 👏👏

style-editor.scss
kjellr: Good thinking. 😄

iseulde:
One thing to be aware of: any non breaking space insertions that get saved accidentally, or any intentional insertions by the user, will also render too wide on the front-end, so this fix might be needed there too.

kjellr:
> Only thing would be to check that the width of this nbsp character set in Georgia is close enough in width to Hoefler that it looks the same

According to my test here, both render at ~15px wide (using our default font size), so I think we're good there: https://codepen.io/kjellr/pen/qQWgNp

Our fallback fonts render _slightly_ slimmer spaces (a couple pixels), so this may cause some very slight issues there. I didn't really notice while testing. However, I'll wait for someone else to confirm too though. 

>  One thing to be aware of: any non breaking space insertions that get saved accidentally, or any intentional insertions by the user, will also render too wide on the front-end, so this fix might be needed there too.

That's a good point. Spaces in the front-end will likely be less noticeable (the only reason this was noticed in the editor as because of the cursor jumping while typing. 😄). But once we're  sure this definitely works for the editor, we can look at rolling it out there too.

jasmussen:
To clarify, I don't feel the width has to be perfect, if it's anything like I think it is (still mobile) then it's a substantial improvement.

The bar to pass is to ensure the writing flow feels right and uninterrupted.

kjellr:
>  uick one: maybe it should be called NonBreakingSpaceOverride (override instead of overwrite)?

Good suggestion. Updated. 👍 

> Also, I need a sanity check on this one. Does the height of the caret change ever so slightly on the space, or am I seeing things? To test, write a sentence, then wait with the caret at the end of the last letter, wait until it's visible, then press space and see if the height changes. It appears as if it becomes 1px taller. If it is, it's fine/acceptable but worth looking at a Georgia alternative that better mimics the height, separately.

I added some guides to my screen to check this, and it doesn't appear to jump for me at least:  

typing

In general, I tested this PR across a few different browsers in Windows, Mac, and iOS. Seems fine in each one. The  only small issue I noticed is that while typing using Hoefler Text, descenders get cut off a little bit:

screen shot 2018-11-02 at 4 03 46 pm

This is resolved once you click out of the field, and is probably related to the line-height. I'll take a closer look on Monday and see if I can sort it out before merging this. 👍

allancole:
Wow. Such a clever fix!!! 👍

kjellr:
> The only small issue I noticed is that while typing using Hoefler Text, descenders get cut off a little bit:
> 
> 
> 
> This is resolved once you click out of the field, and is probably related to the line-height. I'll take a closer look on Monday and see if I can sort it out before merging this. 👍

I don't think it's a line height issue after all: the line height hasn't actually changed — it's 1.8 by default here, whether or not we're defining it where we are in this PR. Commenting our our line height change doesn't have any effect on the issue. 

Instead, my hypothesis is that this has something to do with the fact that Hoefler Text's descenders dip below those of Georgia: 

screen shot 2018-11-05 at 8 30 45 pm

I checked some other potential font options (Times New Roman, Baskerville, etc.), but they all have the same issue — and their descenders all don't dip below the level that Hoefler Text's do. 😞  

In general, still need to dig a little to figure that out... I'm going to focus on other work for a bit, but hope to circle back to this soon.

jasmussen:
I think it's an acceptable trade off, but would definitely be nice if we could find a font alternative that had sufficiently long descenders.

Plan b we could look at a chrome only hack.

iseulde:
One idea to fix the descenders: create a font with just one character for the non breaking space, either taken from Hoefler directly (just take the normal space) or something handcrafted, then base64 encode it (or not) and add it as the source instead of Georgia. I think this is possible. All depends on how long you want to work on this fix of course. Thinking if there might be other solutions...

iseulde:
I don't see this issue in my example btw:

screen shot 2018-11-06 at 09 46 11

https://codepen.io/iseulde/pen/qQWJRo

kjellr:
>  I don't see this issue in my example btw:

Weird — I do see it in Safari 12.0, but it's only noticeable after bumping up the font size a bit, and only when you're typing (not on the text that's there originally):

https://codepen.io/kjellr/pen/pQjjQR

> I think it might be good to ship this as is and then look at additional improvements separately? If need be, I have a license to a font building tool, and can probably create a Hoefler/Georgia hybrid that only contains a space.

My only hesitation is that while this fixes an issue in a single browser (Chrome), on a single operating system (Mac)... it also introduces this clipped text issue everywhere 😕. I'm not 100% sure that's a great tradeoff yet.

jasmussen:
Oh I see it with the j's now. Taking a look.

kjellr:
> Can anyone try and see if this fixes the issue?

Thanks! I pushed a commit that includes this fix. Seems to work during my testing. 👍 Could use another set of eyes to double check though.

style-editor.css
```diff
@@ -10,13 +10,13 @@ Twenty Nineteen Editor Styles
...
- * by inserting a space from the Georgia font instead.
+ * editors. The following font-face override fixes the issue by manually inserting 
+ * a custom font that includes just a Hoefler Text space replacement for that 
+ * character instead.
  */
 @font-face {
   font-family: 'NonBreakingSpaceOverride';
-  src: local("Georgia");
-  unicode-range: U+00A0;
+  src: url("fonts/NonBreakingSpaceOverride.woff2") format("woff2"), url("fonts/NonBreakingSpaceOverride.woff") format("woff");
```
iseulde: Maybe the file can be base64 encoded (it's small), and put inline here to avoid an extra network request. No big deal though. :)

iseulde:
> ☝️ This is a version of the Hoefler font, with the Georgia space character transplanted, and nothing else. Basically Ella's suggestion here #463 (comment)

You could also transplant Hoefler's own space into the non breaking space's spot, which may be even cleaner. I wouldn't affect the space's ability to not break the line, the font is purely for rendering the character itself.

jasmussen:
> You could also transplant Hoefler's own space into the non breaking space's spot, which may be even cleaner. I wouldn't affect the space's ability to not break the line, the font is purely for rendering the character itself.

But isn't the non breaking space spot the same as the space spot? This is getting into advanced territory for me. 

I think the following zip has a base64 encoded version, btw:

[webfontkit-20181107-131226.zip](https://github.com/WordPress/twentynineteen/files/2558628/webfontkit-20181107-131226.zip)

style-editor.css
kjellr: Added via 8701df7!

jasmussen:
We should get this in as soon as possible, the current behavior in master is highly disruptive to the writing flow. Just tested this again and I think it's good to ship and iterate on.

afercia:
I think this needs to be done also for the frontend. Have you seen @swissspidy 's last post on his personal blog? Plenty of `&nbsp;`s there produced by Gutenberg in some way, and plenty of visible, large spaces.

jasmussen:
@afercia See https://github.com/WordPress/twentynineteen/pull/585

jasmussen:
@allancole can you elaborate on the revert?

kjellr:
When Allan made that revert, I'm not sure he realized this had been added to the front-end intentionally. We should re-add.

PR #464, opened 2018-11-02
Disable xdebug to speed up Travis build time

Ismail-elkorchi:
Xdebug is only needed when generating code coverage reports, but, when, not needed, it really slows things down. Disabling it will speed up the build time.

As suggested in : https://johnblackbourn.com/reducing-travis-ci-build-times-for-wordpress-projects/
And in : https://twitter.com/kelunik/status/954242454676475904

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,8 @@ env:
 # e.g. copy database configurations, environment variables, etc.
 # Failures in this section will result in build status 'errored'.
 before_script:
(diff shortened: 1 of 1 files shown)


kjellr:
Thanks, @Ismail-elkorchi!

Ismail-elkorchi:
@kjellr @ntwb Thank you guys for your awesome work !

PR #472, opened 2018-11-03
Add an icon for last.fm

tlxo:
This PR adds an icon for last.fm to the social links menu.


tlxo:
Cool! Forgot to mention that '@tlxo' is also my Wordpress.org username. I don't wanna create another PR just for this one tiny detail, but feel free to add it if you happen to notice this and happen to be updating the contributors list.

PR #481, opened 2018-11-05
Fixed sub menu hover

desi-developer:
I have updated the hover effect for the Submenu to make the parent list item active on hovering the submenu item and also when the user hovers the arrow of the parent nav item in the menu the parent item lose his hover state. I have fixed both issues. 

submenu-hover-fix


allancole:
Closing this since sub-sub-sub-menus no longer open to the left/right of their parent menu. See: #485

PR #479, opened 2018-11-04
Simplify nav submenu markup and improve no-JS behavior

westonruter:
Follow up on #337.

* Eliminate unnecessary duplication of submenu-expand elements (one for desktop and one for mobile).
* Eliminate unnecessary use of link for mobile submenu-expand and menu-item-link-return elements, along with its unnecessary `onclick` attribute.
* Reduce number of CSS rules due to simplified markup.

👉 These changes improve the nav menu experience when JavaScript is turned off in the browser.

Issue #477, opened 2018-11-04
Touchend on new menu will trigger a click to the link at the event position after menu closed


2ndkauboy:
I really like the new menu introduced in #337, but unfortunately it has a random side effect. Once you click the last "back link", the menu will close but trigger the events `mousedown`, `mouseend` and `click` on that exact position, which usually results in clicking the home link (or any other link at that position):

![twentynineteen - navigation touchend issue](https://user-images.githubusercontent.com/1793177/47967852-61fa2380-e062-11e8-929d-9cc7f6064121.gif)

I could confirm this behaviour both in Chrome using the responsive emulation and on a Android smartphone.

PR #485, opened 2018-11-05
Refactor menu UI for better consistency across interaction types

allancole:
**Proposed fixes**
• Supports, tabbed browsing, mouse clicks, and touch interactions
• Removes `:hover` states on mobile devices 
• Fixes the lingering `touchstart` event issue, see #477 
• Cleans up markup in submenus, see #479 
• Adds custom color support for hovered states in the mobile and desktop menu, see #481
• Improve add/remove classes JS to supports IE11 
• Introduces a priority+ UI design, see #313

**To Do**
- [x] Prevent the menu from jumping before the priority+ script fires
- [x] Prevent sub-menus and sub-sub-menus from going off screen, see #398
- [x] Clean up Javascript variable scoping


joyously:
I think it needs the arrow. The indentation alone is not sufficient.

allancole:
> The desktop indicators are needed on “desktop” resolutions for touch devices, yes?

@westonruter Yes, the dropdown arrows are still present for `touchstart` interactions and they’re used to jump back and forth between submenu levels. We’ve only removed them from the the hover/focus-based sub-menus. But since the sub-sub-sub-menus all appear on a single hover, the dropdown icons aren’t clickable and therefore unnecessary for the UI.

allancole:
> I think it needs the arrow. The indentation alone is not sufficient.

@joyously Yeah, I think I agree about the indentation alone not being enough. What if we used a psuedo element to achieve something like this instead?: 

image

Again, these `—`  indicators would only appear on “hovered” sub-menus. The `touch` based menu arrows would remain unchanged.

cc: @westonruter @kjellr

joyously:
That does look better than spaces. Does this method work okay with scrolling (like for a long menu that is tall)?  I assume those with a mouse-wheel can do it easiest.

allancole:
Yes, it scrolls down easiest with the mouse-wheel, and also with the tab-key.

kjellr:
> What if we used a psuedo element to achieve something like this instead?:

Looks good. I added a minor adjustment: matching the font for those en dashes with the face used for the em dash in the header:

screen shot 2018-11-08 at 8 33 07 am

This makes sure the font weight is the same between the two. I also tried an em dash instead of an en dash, but concluded that it was too long for this use case.

kjellr:
This is coming together nicely! 🎉 

I just ran through a bunch of testing in: 

- MacOS 10.14, Chrome 70.0.3538.77
- MacOS 10.14, Safari 12.0
- MacOS 10.14, Firefox 63.0.1
- iOS 12.1, Safari
- Android 6.0.1, Chrome 68.0.3440.85
- Windows 7, IE11

I came across a handful of issues, but most are somewhat minor: 

---

1. Reverse tabbing through the menu still seems to skip submenu items (same issue discussed in #266)

---

2. In IE, I was unable to access submenu items via keyboard navigation at all. 

---

3. On desktop, menu items that follow directly after an item with a submenu appear to be shorter than other menu items:

screen shot 2018-11-09 at 8 48 10 pm

---

4. When child menu item names wrap, they appear below the em dash: 

screen shot 2018-11-09 at 8 47 35 pm

I'd expect these to be intended to align with the rest of the text:

screen shot 2018-11-09 at 8 47 35 pm

_I don't think this is a blocker — we can revise separately if needed._

---

5. I couldn't figure out exactly why/how this happens, but on desktop, in the overflow menu I'm sometimes seeing the some items start appearing squashed alongside others:

screen shot 2018-11-09 at 8 56 03 pm

That example is one of the default menus from the [Theme Unit Test](https://github.com/WPTRT/theme-unit-test), if you want to try to reproduce it that way. Here's another view of that menu, appearing correctly, for reference: http://cld.wthms.co/sqQdwX

This only seems to happen in Chrome. 

---

6. On desktop screens, the usual way to interact with the submenus is via hover. However, the expand buttons are still clickable. This lets you activate more than one submenu at a time:

screen shot 2018-11-09 at 9 09 01 pm

Not totally ideal, but it's easy to get out of at  least. _I think this could be addressed in a separate PR if necessary._

---

7. When the touch menu is active, it'd be cool to try adusting the top-level expand icons to point right instead of down (or be replaced with a plus icon or something?), [as per the suggestion here](
https://github.com/WordPress/twentynineteen/issues/143#issuecomment-431398156). Not a blocker though. 

---

8. For the mobile menu, we have nice slide-in animations, but we do not have slide-out animations when the menu is closed. _Not a blocker, and could be addressed later on._

---

9. On mobile, I noticed that if I tap on a top-level item that's mapped to `#`, I can activate the dropdown menu instead of the normal mobile menu. Here's a recreation via the iOS simulator so you can see where I'm tapping:

iphone

---

10. On iOS, when I close a top-level full-screen menu, I see a quick flash of the associated dropdown menu. _Not a blocker — could be addressed in a separate PR._

---

11. On iOS, if I tried to visit a menu item that had children, I had to tap twice on the title to get there. The first tap had no effect. 

screenrecording_11-09-2018 21-39-33

---

12. In IE11, the priority+ menu is cut off on the right side: 

virtualbox_ie11 - win7_1_09_11_2018_21_53_49

---

13. Also in IE11, I'm seeing some non-ideal line breaks:

virtualbox_ie11 - win7_1_09_11_2018_21_54_55

allancole:
@kjellr The above commits address notes 2, 5, 9, 10, 11, 12, and 13 from above. I’m going to merge this in now, so we have something for Beta4, and then do new PRs for the other issues as noted below.

> 1. Reverse tabbing through the menu still seems to skip submenu items (same issue discussed in #266)

Will fix this in another PR

> 3. On desktop, menu items that follow directly after an item with a submenu appear to be shorter than other menu items:

Couldn’t replicate this one. Which browser is this happening in?

> 4. When child menu item names wrap, they appear below the em dash:

Will fix this in another PR

> 6. On desktop screens, the usual way to interact with the submenus is via hover. However, the expand buttons are still clickable. This lets you activate more than one submenu at a time:

This comes from a conflict between `:hover` and `:focus-within` states. Both are necessary to allow keyboard browsing and there doesn’t appear a way to fix this without effecting other behaviors. May need to leave this in.

> 7. When the touch menu is active, it'd be cool to try adusting the top-level expand icons to point right instead of down (or be replaced with a plus icon or something?), as per the suggestion here. Not a blocker though.

Might not be possible to swap the down arrow SVG for a `+` without adding more markup here—not sure it’s worth it. 

> 8. For the mobile menu, we have nice slide-in animations, but we do not have slide-out animations when the menu is closed. Not a blocker, and could be addressed later on.

May try to fix this in another PR.

PR #488, opened 2018-11-05
Syntax error in functions.php

briannaorg:
Received “syntax error, unexpected ')' in
/var/www/html/wp-content/themes/twentynineteen/functions.php on line
105” on first load of site after merging from master. Added explicit
value to fix.


briannaorg:
Merge issue on my end - closing PR

Issue #487, opened 2018-11-05
3 Column Gallery Widget display issues


allancole:
The right padding is removed from the last image in a 3-column gallery which makes it appear larger than the other gallery images. The CSS should be revised to use an `:nth` selector for every 3rd image when a user selects a 3-column gallery. The same should happen for all other column count settings (1-9 columns). 

Screenshot: 

![image](https://user-images.githubusercontent.com/709581/48020225-63c7f380-e103-11e8-917d-01f0fba10226.png)

briannaorg:
I'll take this one if it's unassigned

PR #489, opened 2018-11-05
Fix 487

briannaorg:
Fix for #487. Remove :last-of-type designation, match item padding between front end
and editor.

I looked at the history of _galleries.scss and it’s like a conversation between @allancole and I. :) At any rate, rather than designating a right side padding by column number using nth-of-type, I tested what would happen if we omitted it altogether. Then checked the editor styles and found that they handled the item padding better than we were on the front end, so I matched that by defining the horizontal padding for all items instead. Tested the shortcode in the Theme Unit Test Data and also added it to a Classic Editor block. Device tested.


briannaorg:
thanks @allancole - this highlighted that my test images are all displayed with captions, because I didn't notice the vertical spacing. I tested your fix on #490 and it looks great over here, thanks again for the detailed response!

allancole:
Gotcha! I’ll close this for now in favor of #490, but I’ll try tag take a few more “first-commit” issues, and actually not fix them on my end next time :-)

PR #491, opened 2018-11-05
Remove duplicate mention of my username in the CONTRIBUTORS.md file

Ismail-elkorchi:
My username appears twice in the list. So this PR removes the last occurrence.


allancole:
Woops, that was my mistake. Thanks for contributing — twice! :-)

PR #502, opened 2018-11-07
Update front end margins to match the editor on desktop screens

### LINKED ISSUE #236: Sync up margins, line-length between frontend and editor
#22 introduced greater visual consistency between the front and back end. However, it can be improved. We should take another pass at making sure both the editor and front end render line-lengths and margins similarly across common breakpoints. 

**Current state:**

artboard 0

artboard 1

artboard 2

kjellr:
Partially resolves #236

Moves our tablet + desktop margins around a bit so that there's greater alignment between our front and back end. Basically: if you're in full-screen mode in Gutenberg (on a screen above `768px` wide), you should see _nearly_ the exact same margins/grid for your post content when you switch to the front end. 

- This is based on [the margins/spacing that are built into `style-editor.css`](https://github.com/WordPress/twentynineteen/blob/fac38ef358fc7ab6794a5ef7a516a0b530f28c53/style-editor.scss#L23-L53), so starts by adding `10%` left/right margins to `.entry-content`. There are a few extra pixels added to sync up with Gutenberg's block/container padding. From there, it just modifies the existing margins + max-widths of blocks (and other non-block post elements) so that they fit the new grid.
- When possible, I switched to using mixins/variables for some of our common body content max-width specs. 
- While I was adjusting the comment margins, I also made it so that our comment reply links don't disappear on small screens. 😉

I ran through my entire test site (including all the theme unit test posts), and things looked fairly solid. 

*In order to fully resolve #236, we'll need another PR that syncs up the margins between `600px` and `768px` viewports.

**GIF:**

margins

### DIFF
--- a/inc/color-patterns.php
+++ b/inc/color-patterns.php
@@ -56,12 +56,12 @@ function twentynineteen_custom_colors_css() {
 		.sticky-post,
 		.entry-content .wp-block-button .wp-block-button__link,
 		.button, button, input[type="button"], input[type="reset"], input[type="submit"],
-		.entry-content > .has-primary-background-color,
-		.entry-content > *[class^="wp-block-"].has-primary-background-color,
-		.entry-content > *[class^="wp-block-"] .has-primary-background-color,
-		.entry-content > *[class^="wp-block-"].is-style-solid-color,
-		.entry-content > *[class^="wp-block-"].is-style-solid-color .has-primary-background-color,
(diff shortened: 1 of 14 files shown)


kjellr:
🤔 Ah, I bet that's due to adding the `.entry` wrapper to our block styles. I can fix. 👍

kjellr:
@allancole Should be all set now. Mind re-testing?

PR #507, opened 2018-11-07
🐛 FIX: Markdown Lint Issues

ahmadawais:
This PR fixes two markdown lint issues with the `README.md` file.

- [x] <kbd>2</kbd> x Multiple consecutive blank lines [Expected: 1; Actual: 2]
- [x] <kbd>1</kbd> x Lists should be surrounded by blank lines
- [x] <kbd>6</kbd> x Empty whitespace characters at line end

Peace! ✌️

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,14 +2,13 @@
 
 [![Build Status](https://travis-ci.org/WordPress/twentynineteen.svg?branch=master)](https://travis-ci.org/WordPress/twentynineteen)
 
-**Contributors:** the WordPress team  
-**Requires at least:** WordPress 4.9.6  
-**Tested up to:** WordPress 4.9.8  
-**Version:** 1.0  
-**License:** GPLv2 or later  
-**License URI:** http://www.gnu.org/licenses/gpl-2.0.html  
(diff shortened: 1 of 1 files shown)


ahmadawais:
@crunnells Awesome, looking forward to this PR merged! 💯

allancole:
Looks good 👍

allancole:
Good catch @ocean90 👍 
Rolling that back now…

ahmadawais:
Was not aware of two white spaces == line break rule. Seems pretty odd to do that. Also, the white spaces in this case are rendering <br /> elements which _might be_ semantically wrong. Line breaks should be created with styling and not with elements. HTML elements are not meant to be used as styles. 

May I suggest using ul/li instead? i.e.

- **Contributors:** the WordPress team
- **Requires at least:** WordPress 4.9.6
- **Tested up to:** WordPress 4.9.8
- **Version:** 1.0
- **License:** GPLv2 or later
- **License URI:** http://www.gnu.org/licenses/gpl-2.0.html

↑ instead of `  ` two white spaces.

PR #509, opened 2018-11-07
🐛 FIX: inline documentation

ahmadawais:
This PR fixes several inline docs issues as per the [WordPress inline docs standards](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/)

- [x] 🐛 FIX: Inline comments should end with a period
- [x] 🐛 FIX: `require`/`include` are not functions, they are language constructs - [no parentheses!](https://tfrommen.de/on-php-language-constructs-and-parentheses/)
- [x] 🐛 FIX: Function definition should use a period at the end
- [x]  📖 DOC: : New Inline documentation

Peace! ✌️


ahmadawais:
Thabks. I can remove that if you want. But for me it solves a red squiggly line due to not following the inline docs standards via PHPCS.

allancole:
Looks good as is. Gonna go ahead and merge

Issue #216, opened 2018-10-20
What sort of starter content would work well for this design?


joyously:
Is this theme going to supply starter content? 
Will it have blocks? Does core support that yet, or is it not necessary for core to do anything different?

One thing it could have is the menu, and the menu could have privacy link.

celloexpressions:
Starter content is a relatively new feature in core, dating to WordPress 4.7. As a relatively new feature, and given that bundled themes should be "state of the art," it seems like a mandatory element of Twenty Nineteen. We should also consider the goal of starter content to improve the new user experience with fresh installations of WordPress. Leaving this out of Twenty Nineteen would be a regression relative to the current fresh-install experience with Twenty Seventeen.

This would be a great area for someone with experience building the types of sites that this theme is intended for to contribute.

LittleBigThing:
As proposed during one of the Slack chats, I have looked into adding starter content to the theme.

It is explained above why starter content is useful, but I think that for this theme in specific, it could additionally highlight the new block editor's features. It could give a first impression of what can be achieved, but only some basics seem sufficient. And also how this can be achieved using the editor. Once the starter content is saved and a page containing it is edited by the user, (s)he could learn the basics about using the editor. It is certainly an enhancement.

I'll added a PR with a first very rough try for starter content: #510.

Standard, [core provided starter content](https://core.trac.wordpress.org/browser/trunk/src/wp-includes/theme.php?rev=39373#L1910) is not recognised as a block. Starter content needs to contain Gutenberg markup. These are the HTML comments added before and after blocks by Gutenberg, for example `<!-- wp:paragraph -->`, including attributes if present (`<!-- wp:paragraph {"dropCap":true} -->`). Gutenberg will recognise such content as blocks, and they will be nicely editable. To create such starter content, we need to add the blocks we wish to have in the Gutenberg editor, changing to 'code editor' and copying the markup to add it is as starter content. To make it translatable, additional translation functionality (`_x()`) need to be wrapped around text strings.

**What can we add?**
- create a front page using different blocks; for example, columns, a wide-aligned image, a cool blockquote, some color settings, ...
- theme_mods: a page as front page, a blog page ('blog')
- a contact or an about page: core provides these, but they should be updated with some Gutenberg markup to have them opened as blocks and not as a classic block.
- ...

**What do we need for this?**
- design to decide which blocks we use
- content: this can be a bottleneck to, also with regard to translation

[...]

PR #510, opened 2018-11-07
add starter content

LittleBigThing:
Related to #216 

A first, rough take on adding Gutenberg compatible starter content to the theme


joyously:
Can you change the content out to something that's actually useful for a new site? I know you want to introduce the theme, but it will be the default theme for awhile, and new sites don't need that. They need a nice way to start their site.

LittleBigThing:
@joyously Of course, but as I mentioned above, I would love to have suggestions on the content we’ll use, blocks as well as text.

Do we go for a blog example or for for a business site? Maybe there are clearer directions for this already.

joyously:
There aren't. We just have to make it up as we go. That's why I asked for what this theme was targeting, and the answer I got was basically "everything" or "highlight GB".

For starter content, it makes more sense to create Pages and set a home page than it does to create Posts and show the blog on home, even though WP is blogging software. The user should not have to "undo" the starter content. It should be more like fill in the blanks.

allancole:
I haven’t tested this yet @LittleBigThing, but thanks for getting this started! 

Regarding what the content should be, I think we might already have some good starter content we can borrow from our early mocks. Maybe these could work: 

A Blog **Post**: https://cloudup.com/ijfvuiEe5ne
A Business **Page**: https://cloudup.com/i9tJJqeq47A

I also have this content built out with Gutenberg Blocks on my local testing site. I could send that to you as an WP export file or just share the raw Gutenberg markup, if you want. 

> The user should not have to "undo" the starter content. It should be more like fill in the blanks.

Agreed and with Gutenberg, it’s super easy to swap out images/text.

LittleBigThing:
@allancole cool, welcome to send me that raw content in one way or another so I can “quickly” update the PR.

allancole:
Here goes it @LittleBigThing :-) 

**Blog Post Content**
https://gist.github.com/allancole/4131162468ccc5da8dbca24478cd52e4

**Business Content**
https://gist.github.com/allancole/2f69f6a9ad1f9fcc891ff269f00401ec

**Image Assets**
https://www.dropbox.com/s/54exvar23ci2g4h/Assets.zip?dl=1

WarenGonzaga:
I have suggestion, can we use "Gutenberg Ipsum" a dummy text for the starter content? Something link an alternative for Lorem Ipsum. Reference: http://mentalfloss.com/article/61003/10-funnier-alternatives-lorem-ipsum

LittleBigThing:
So a late update, if not too late.

I was stuck at adding images contained in the post content, such as a background image for a Cover block (see [proposed starter content design](https://cloudup.com/i9tJJqeq47A)).

As an example, the cover block has this HTML markup:

```
<!-- wp:cover {"url":"http://gutenberg.development/wp-content/uploads/2018/10/office-space.png","align":"full","contentAlign":"left","id":3370} -->
<div class="wp-block-cover has-background-dim has-left-content alignfull" style="background-image:url(http://gutenberg.development/wp-content/uploads/2018/10/office-space.png)"><p class="wp-block-cover-text">Digital strategies for unique small businesses</p></div>
<!-- /wp:cover -->
```

So we need a URL and an id to have this image in a cover block. It seems that we can insert a URL, but I am not sure we can get an attachment ID there. But is it needed? Can someone test it if this is in any way an interesting addition so last minute.

Another thing I noticed is that starter content for a page/post needs to have a title (so I had to add something). In the proposed design, there is no page title, but a cover image with text (in a paragraph) that functions as page title. In Gutenberg there is no way of setting the text on a cover image as a heading, which we would probably need. 

**We could slightly adjust the design and use a featured image with page title instead of the cover block?**

LittleBigThing:
To summarize what's added here as starter content:

- a custom front page built with Gutenberg
- a Blog page to display posts
- an About page (core starter content), with a featured image
- a Contact page
- 2 attachments: a landscape and a portrait placeholder image
- 2 menus: a main menu containing the above pages in that order and a social menu

Placeholder images were used instead of photographs for now.

LittleBigThing:
Hi, I've just updated the PR to the content on the [demo site's front page](https://2019.wordpress.net/). Except for the images, they are still placeholders, since I am not sure about the sources and the licenses.

Further, I am not sure how to port PRs over to Trac or convert them to patches. I can start by creating a ticket.

allancole:
Hi @LittleBigThing, I don’t see an issue for this on Trac, so I’ll go ahead and take care of it and work out a patch for you. 👍

allancole:
I've moved this PR to Trac as a patch here:

https://core.trac.wordpress.org/ticket/45927

I'll close this issue for now, and we can continue discussion over in that thread. Thanks!

PR #315, opened 2018-10-25
Only show dash in header if site title and tagline are available

nielslange:
Fixes #310

<table>
<tr>
<td>Before:

310 - before site title tagline</td>
<td>After:

310 - after site title tagline</td>
</tr>
</table>

<table>
<tr>
<td>Before:

310 - before only site title</td>
<td>After:

310 - after only site title</td>
</tr>
</table>

<table>
<tr>
<td>Before:

310 - before only tagline</td>
<td>After:

310 - after only tagline</td>
</tr>
</table>

PR #511, opened 2018-11-07
Change the way the EM Dash is added to tagline

### LINKED ISSUE #451: "-" character is shown when the Tagline is empty
If the title is filled out but the tagline is left empty, we can see the "-" character after the title in the front-end:

screen shot 2018-11-01 at 19 26 47

This seems to appear in `site-description::before`

screen shot 2018-11-01 at 19 29 08

ianbelanger79:
This PR is a possible solution to the problem of the EM Dash character showing in the customizer, even when the Tagline is empty. It works, but may need some refinement.

Fixes #451

### DIFF
--- a/style.css
+++ b/style.css
@@ -1744,11 +1744,6 @@ body.page .main-navigation {
   }
 }
 
-.site-title:not(:empty) + .site-description:not(:empty):before {
-  content: "\2014";
-  margin: 0 .2em;
-}
-
 .site-description {
   display: inline;
   color: #767676;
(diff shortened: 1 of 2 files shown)


nielslange:
Thanks for your contribution, @ianbelanger79! However, in addition to what @enodekciw mentioned, in #129 the dash had been removed from the HTML and added to CSS and in #315 we've ensured that the dash is only visible, when both the site title and the tagline are available. Your PR would show the dash, when the tagline is available, even when no site title is available. 

Can you think of any other way to target the dash in the customizer? Or shall I take over again?

allancole:
Seems like the only way to make this work as expected is to add some Javascript to `js/customizer.js` that conditionally overrides the psuedo element.

kjellr:
Since this particular approach doesn’t seem to be viable as per the notes above, I’ll go ahead and close this PR for now.

PR #512, opened 2018-11-08
Update the colorscheme function to set the hue correctly 

### LINKED ISSUE #468: Custom color indicator does not match the preview at first
I thought this was fixed.
Choosing `Default` color changes the preview to blue.
Choosing `Custom` color does not change the preview, but the hue indicator is on the color last chosen (as I would expect). Selecting a different color does update the preview.
The problem is that the hue indicator does not match the preview when it is first shown.

custom-hue-2018-11-02 21-44

brentswisher:
The color scheme customizer function currently takes in a string, either "default" or "custom" as it's 'to' argument, but tries to set the hue to that actual value. This PR updates that to grab the current custom hue or use the default hue instead based on the to value passed in, fixes #468

Ideally, we could pull the default hue from somewhere without hardcoding it, but I couldn't find anywhere that it was stored. I thought "wp.customize.settings.values.colorscheme_primary_hue" might work, but that was just the last saved value, not the true default, so I hard coded it.

(Wordpress.org username is also brentswisher)

rwx2ioaeo3

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -67,3 +67,4 @@ Version 1.0:
 | @whizbangik | @imonly_ik |
 | @yingles | @yingling017 |
 | @tlxo | @tlxo |
+| @brentswisher | |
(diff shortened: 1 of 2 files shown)


kjellr:
Thanks @brentswisher! I can confirm that this works. 👍 

cc @allancole for a review too.

Issue #45, opened 2018-10-16
Add Footer Menu


00travelgirl00:
Add a Footer Menu Position, do make it easier for users to integrate a privacy menu page. In certain countries websites have to have an "Impressum" and privacy page on the websites. The footer is the mst common space to integrade it.

melchoyce:
Could this be accomplished by styling a menu widget to fit well within the footer widget area? (#14)

audrasjb:
Oups sorry @melchoyce for the -1 emoji, I am on my mobile and not able to change it for the moment :(((

I agree your solution can work but I'm also wondering if that wouldn't be more easily discorable in menu section since you don't have to 1) create a menu then 2) add it in a widget :)

audrasjb:
Oups sorry @melchoyce for the -1 emoji, I am on my mobile and not able to change it for the moment :(((

I agree your solution can work but I'm also wondering if that wouldn't be more easily discorable in menu section since you don't have to 1) create a menu then 2) add it in a widget :)

nielslange:
@audrasjb In the WP.com support we received several requests on how to add additional links behind the “Privacy Policy” link. Thus, I would say having a footer menu, in addition to the footer widget section, make sense. I’m happy to implement this later and create a PR.

Zodiac1978:
Just wanted to add this issue as well. This is the most added little feature in my child themes for clients. +1 from me.

nielslange:
Closed as fixed in #514.

PR #514, opened 2018-11-08
#45: Added Footer Menu

youthkee:
Added Footer Menu for additional links behind the "Privacy Policy" link.
Related to #45 .

twentynineteen-add-footer-menu


kjellr:
This seems okay to include, but I noticed that it does not support submenus. Is that a problem? And if it is, how should those menus work?

It seems overkill to build new flyout/dropdown menu functionality down in the footer. My gut says that we should take the simple route and expose submenu items with no indentation or anything in this case: all items, whether top level or children would appear the same. I'm interested to hear thoughts from @allancole and others though.

allancole:
@kjellr yeah, I agree. This menu should not support sub-menus at all. 

> My gut says that we should take the simple route and expose submenu items with no indentation or anything in this case: all items, whether top level or children would appear the same. 

Pretty sure we can accomplish this with some simple CSS. But we should also set the depth of the menu to `'depth' => 1`, so sub-menus are totally unsupported.

kjellr:
> Pretty sure we can accomplish this with some simple CSS. But we should also set the depth of the menu to 'depth' => 1, so sub-menus are totally unsupported.

That sounds great. If we add that ` 'depth' => 1` to the menu, we don't need the extra CSS. @youthkee would you mind making that small update? Then this should be good to go. 🎉

PR #523, opened 2018-11-09
Don’t escape translations

WPprodigy:
The opposite of https://github.com/WordPress/twentynineteen/pull/285. If we aren't going to escape every translation, then I suppose we should essentially escape none (that hurts to say).

Removes `esc_html_e()`, `esc_html__()`, and `esc_html_x()` and replace with the non-escaping version.

I left one in use as it is also escaping a variable: https://github.com/WordPress/twentynineteen/blob/5e8fa546d5dacfa44fa33ccf152b556992bc53de/inc/template-functions.php#L90

Also removed a stray space in the markup in `image.php` that I came across.
_________

Just noting that I'm still quite opposed to this. Three reasons why I believe this is the wrong path (sorry, didn't know where else to post 🙂): 

1) Translation plugins are widely used and allow users to alter text strings from the admin. So you essentially have an admin setting that is then directly output on the frontend. In every other case, this would 100% need to be escaped.

2) Translations come from a separate and outside "database". I think the decision to not escape translations was made before the current translation system on wp.org. I'm not very experienced with it, but from what I know this could be an attack vector where the attacker only needs access to (or have their own) account that has permissions needed to approve translations. Only needs to happen once, and could have maximum impact in a theme like this.

3) It's much harder to add this escaping at a future date after release. If there were to be future reasons where we needed to go back and add escaping (perhaps the above 2 reasons or something else is revealed), then this is hard to do while keeping full backwards-compatibility. Inevitably some user somewhere is going to have HTML in a translation and will be upset when it stops working.

And lastly, to counter one of the counterpoints brought up in https://core.trac.wordpress.org/ticket/30724:

> That's simple. If we don't trust translations anymore we should do it directly in the function. `function __() { return esc_html(….`.

There are times where translations need to include html, seen recently in this project here https://github.com/WordPress/twentynineteen/pull/274. In these cases, you need `__()` as it is, and then `wp_kses()` around it to allow whichever tags you are using.

### DIFF
--- a/footer.php
+++ b/footer.php
@@ -25,7 +25,7 @@
 			<a href="<?php echo esc_url( __( 'https://wordpress.org/', 'twentynineteen' ) ); ?>" class="imprint">
 				<?php
 				/* translators: %s: WordPress. */
-				printf( esc_html__( 'Proudly powered by %s.', 'twentynineteen' ), 'WordPress' );
+				printf( __( 'Proudly powered by %s.', 'twentynineteen' ), 'WordPress' );
 				?>
 			</a>
 			<?php
--- a/comments.php
+++ b/comments.php
@@ -29,9 +29,9 @@ if ( post_password_required() ) {
 		<?php
 			if ( comments_open() ) {
 				if ( have_comments() ) {
-					esc_html_e( 'Join the Conversation', 'twentynineteen' );
+					_e( 'Join the Conversation', 'twentynineteen' );
 				} else {
-					esc_html_e( 'Leave a comment', 'twentynineteen' );
+					_e( 'Leave a comment', 'twentynineteen' );
 				}
 			} else {
 				$comments_number = get_comments_number();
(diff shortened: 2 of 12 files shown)


footer.php
```diff
@@ -25,7 +25,7 @@
 			<a href="<?php echo esc_url( __( 'https://wordpress.org/', 'twentynineteen' ) ); ?>" class="imprint">
 				<?php
 				/* translators: %s: WordPress. */
-				printf( esc_html__( 'Proudly powered by %s.', 'twentynineteen' ), 'WordPress' );
+				printf( __( 'Proudly powered by %s.', 'twentynineteen' ), 'WordPress' );
```
grappler: We could just change the string to `_e( 'Proudly powered by WordPress.', 'twentynineteen' )`
```suggestion
				_e( 'Proudly powered by WordPress.', 'twentynineteen' );
```

WPprodigy: There is precedent to this: https://github.com/WordPress/twentyseventeen/blob/master/components/footer/site-info.php#L13. I suppose to avoid "WordPress" from being changed / translated.

PR #526, opened 2018-11-09
Fix additional issues with columns block.

jasmussen:
The previous fix to nested blocks in general did not sufficiently address the cascading of properties. Recent refactors to the columns block in Gutenberg upstream made this visible.

This PR fixes those issues by increasing the specificity of a few of the rules that would also target any nested blocks.

Before:

before

After:

after

### DIFF
--- a/style-editor.scss
+++ b/style-editor.scss
@@ -19,7 +19,7 @@ body {
 	@include media(mobile) {
 		overflow-x: hidden;
 
-		.editor-block-list__layout,
+		:not(.editor-inner-blocks) > .editor-block-list__layout, // Target only the top level layout element, or nested blocks will also be affected.
 		.editor-post-title {
 			padding-left: 0;
 			padding-right: 0;
@@ -33,7 +33,7 @@ body {
 
 	@include media(tablet) {
 
-		.editor-block-list__layout,
+		:not(.editor-inner-blocks) > .editor-block-list__layout, // Target only the top level layout element, or nested blocks will also be affected.
 		.editor-post-title {
 			padding-left: 46px;
 			padding-right: 46px;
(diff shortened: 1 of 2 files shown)


style-editor.scss
```diff
@@ -19,7 +19,7 @@ body {
 	@include media(mobile) {
 		overflow-x: hidden;
 
-		.editor-block-list__layout,
+		:not(.editor-inner-blocks) > .editor-block-list__layout, // Target only the top level layout element, or nested blocks will also be affected.
```
jasmussen: I am aware that this rule is relatively exotic, but it's the shortest and most readable way I could write it.

To explain what's going on — any _nested_ block will always have both `editor-block-list__layout` and `editor-inner-blocks` as parent elements. Which means any CSS rules we apply to `editor-block-list__layout` will cascade into nested blocks. 

By writing this rule, `:not(.editor-inner-blocks) > .editor-block-list__layout`, we are targeting any instance of `editor-block-list__layout` that is _not_ a descendant of `editor-inner-blocks`. Which incidentally means, only the top level one is matched. 

I'd love for a simpler rule, but it absolutely has to be solid enough to target only the to level element.

style-editor.scss
```diff
@@ -82,6 +82,7 @@ body {
 		width: calc(6 * (100vw / 12 ));
 	}
 
+	// Only the top level blocks need specific widths, therefore override for every nested block.
```
jasmussen: See also https://github.com/WordPress/twentynineteen/pull/526/files#r232209089 — I wish there was a different way we could write this rule.

Right now we are setting widths on every `wp-block`, and then _unsetting them again_ on every nested block. It would be more solid if we could target only the top level `.wp-block`, because I could easily imagine a Gutenblock that is not meant to be 100% wide.

style-editor.scss
allancole: Yeah, this unfortunate, but will have to do for now.

Issue #460, opened 2018-11-02
Consider using a container to wrap content


psaikali:
Hi!

I really like the idea of NOT having a container around the content markup, and solely depending on margins.
But any plugin that will output extra content (i.e. before or after post content) will not be able to center it without a custom class. 
Or even creating a child theme and adding some extra data here and there will require adding custom CSS to center it.

A simple example below using Jetpack and its Related section. 
![capture 2018-11-02 a 12 05 57](https://user-images.githubusercontent.com/1606655/47912285-607cff80-de98-11e8-81be-fd8dbd98eae5.jpg)

What about wrapping the main content div in a container?

joyously:
+1 to this
If the container has padding, no margins are needed, except for using negative margins for alignwide or alignfull.
It is too fragile the current way.

kjellr:
#502 was merged yesterday, and added margins to many common content wrappers (`entry-content`, `entry-summary`, etc.). This helps a lot, and fixes the Jetpack example mentioned above. 

In order to fully address this, we still need to implement a similar fix for small screens. (That PR only addressed breakpoints over `768px`). So I'm leaving this issue open until that's addressed. 👍

PR #534, opened 2018-11-09
On small screens, move left/right margins to parent wrappers instead, minor comments area fixes.

kjellr:
A counterpart to the recent margins updates in #502. Fixes the issue reported in #460. 

Should generally have no effect on the visual appearance of default theme elements. But this will make sure that when a plugin adds an element to entry-content, the element has the correct `1rem` left/right margins on smaller screens. 

This also applies the same treatment to our `comment-area` children, instead of styling them each individually using a `*` rule.

Example with Jetpack related posts: 

**Before:**
screen shot 2018-11-09 at 1 21 56 pm

**After:**
screen shot 2018-11-09 at 1 21 45 pm

### DIFF
--- a/inc/color-patterns.php
+++ b/inc/color-patterns.php
@@ -87,8 +87,7 @@ function twentynineteen_custom_colors_css() {
 		.entry .entry-footer a:hover,
 		.entry .entry-content .more-link:hover,
 		.main-navigation .main-menu > li > a + svg,
-		.comment-navigation .nav-previous a:hover,
-		.comment-navigation .nav-next a:hover,
+		.comment .comment-metadata > a:hover,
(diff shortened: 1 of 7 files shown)


allancole:
Found some minor color inconsistencies in comment links. Fixed them in the above commit.

kjellr:
Looks good! I made a couple more small adjustments to the comments area on tablet (and larger) screens:

- Adjusted the left margins to avoid having the avatars hang off the left side.
- Fell back to the small-screen treatment of the author meta (stacked name/date info, instead of all on one line). I was sometimes seeing weird text-wrapping issues when authors had long names, and doing this gives helps prevent that:

screen shot 2018-11-09 at 7 54 23 pm

This should be good to go. 👍

PR #538, opened 2018-11-10
Update widgets to use sans-serif fonts for all list-style widgets

allancole:
I noticed some font inconsistencies between lists-style widgets. 

Screenshot before:
image

This PR makes any list-style widgets use the same fonts and list styling:

Screenshot after: 
image

I also added dashes to the nested list styles so they match the menu’s nested styles: 

image


allancole:
Made a small change to the post count text-color in list widgets to match the frontend.

**Before**: 
image

**After**:
image

PR #540, opened 2018-11-10
Fixes for IE 11 on sub menu hover

2ndkauboy:
This PR will fix the two IE 11 issues 12. and 13. as well as another issue, were the hover on the "priority+" menu did not worked as reaching the menue (even with a quick mouse movement) was not possible:

**Before:**

twentynineteen - plus navigation ie11

**After:**

twentynineteen - plus navigation ie11 - fixed

See #485

Username: Kau-Boy

### DIFF
--- a/sass/navigation/_menu-main-navigation.scss
+++ b/sass/navigation/_menu-main-navigation.scss
@@ -43,6 +43,7 @@
 
 			&.menu-item-has-children {
 
+				display: inline-block;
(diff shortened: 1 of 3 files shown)


2ndkauboy:
I have tested my changes in IE 11, Chrome, Firefox and Edge (which had the same issues 12. and 13. as IE 11).

PR #545, opened 2018-11-10
Restrict featured images to only post and page post-types

allancole:
This ensures that Products, Portfolios, Testimonials and any other custom-post-types do not display featured images by default. 

A theme/plugin author can always use the `twentynineteen_can_show_post_thumbnail` filter to define their own conditional restrictions on when the featured header image should be displayed. 

WooCommerce as an example, utilizes featured images in a _non-standard_ way and we don’t want the theme to assume how it (or any plugin) might want to display featured images.

### DIFF
--- a/inc/template-functions.php
+++ b/inc/template-functions.php
@@ -99,7 +99,7 @@ add_filter( 'get_the_archive_title', 'twentynineteen_get_the_archive_title' );
  * Determines if post thumbnail can be displayed.
  */
 function twentynineteen_can_show_post_thumbnail() {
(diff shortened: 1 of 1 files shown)


grappler:
I am not sure I agree with the premise. This would mean that any plugin or CPT that wants their featured images to be displayed needs to add special support for the Twenty Nineteen theme? 

I feel that other plugins would be penalised because WooCommerce utilizes featured images in a **non-standard** way.

allancole:
Good point, @grappler. This change is too aggressive and should work the other way around so that if you’re doing a non-standard thing with featured images, then you’ll need to adjust the conditional filter. Closing this one 👍

PR #546, opened 2018-11-10
General code and documentation improvements

grappler:
While going through the code I found a few places where the code or the documentation could be improved. For each point that I found I made a separate commit.

There should not be anything controversial.

Happy to answer anything if the commit messages are not clear.

### DIFF
--- a/inc/template-functions.php
+++ b/inc/template-functions.php
@@ -415,8 +415,7 @@ function twentynineteen_hsl_hex( $h, $s, $l, $to_hex = true ) {
 
 		return "#$r$g$b";
 
-	} else {
-
-		return "rgb($r, $g, $b)";
 	}
+
+	return "rgb($r, $g, $b)";
 }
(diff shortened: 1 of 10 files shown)


inc/template-functions.php
```diff
@@ -106,10 +106,7 @@ function twentynineteen_can_show_post_thumbnail() {
  * Returns true if image filters are enabled on the theme options.
  */
 function twentynineteen_image_filters_enabled() {
-	if ( 'inactive' === get_theme_mod( 'image_filter' ) ) {
-		return false;
-	}
-	return true;
+	return 'inactive' === get_theme_mod( 'image_filter' );
```
WPprodigy: This seems flipped to me. It'll now return true when it should be false.

grappler: Yes, you are right. It came to me after I had created the PR and closed the computer.

grappler:
@allancole I have fixed the code conflicts again. Is there a reason that the PR has not been reviewed yet.

I am happy to move any commits to their own PR if you feel they need to be discussed further.

Issue #558, opened 2018-11-12
Elipsis Icon shows in menu on all Screen sizes


pixolin:
No matter how long the menu is, I see an ellipsis icon (svg) (circle with three dots) behind the last menu entry, viewing the web page on my desktop monitor:

<img width="414" alt="screenshot_352" src="https://user-images.githubusercontent.com/1968542/48349733-566fb380-e685-11e8-9c00-17494e510a51.png">

If I edit the menu in the customizer, same icon shows up twice.

<img width="679" alt="screenshot_351" src="https://user-images.githubusercontent.com/1968542/48349755-6ab3b080-e685-11e8-8cb9-e822681345e1.png">

kjellr:
Fixed via #589 👍

PR #568, opened 2018-11-12
Fix priority+ menu when the customizer refreshes

allancole:
Addresses #558 

This is an in-progress temporary fix that simply removes the priority+ behavior from the main menu in the customizer. The priority+ script is written in vanilla Javascript while the customizer itself relies on jQuery. Usually this would be fine, but [vanilla JS cannot listen for events created by jQuery](https://stackoverflow.com/questions/40915156/listen-for-jquery-event-with-vanilla-js), so when the menu gets refreshed there’s no way to re-run the script.

Instead forcing a jQuery dependency on the frontend, I think the next best solution is to add a new customizer-only jQuery version of the priority+ script and only run it in the customizer.


allancole:
Closing this in favor of PR #589 which is a better solution.

Issue #564, opened 2018-11-12
Twenty Nineteen breaks the Gutenberg meta boxes area


afercia:
Tested on WP 5.0 (happens also on 4.9.8 with Twenty Nineteen added cloning the repo) and Yoast SEO activated:

The meta boxes area overlaps and almost completely hides the content area:

<img width="1279" alt="screenshot 2018-11-12 at 17 03 22" src="https://user-images.githubusercontent.com/1682452/48360022-86778080-e69e-11e8-97a1-fd787d45ed4a.png">

Activate Twenty Seventeen to double check: everything looks fine.

Suspected CSS rule:
```
.editor-styles-wrapper {
    overflow-x: hidden;
}
```

PR #575, opened 2018-11-13
Fix so sticky toolbar works again.

jasmussen:
When you are not using the "unified toolbar" option, toolbars scroll with you down the page if content inside it is long.

A recent addition broke that, so they were fixed to the top left of the block. This PR removes the rule that broke it. To explain what's going on, the overflow rule makes the `position: sticky;` not behave as intended. 

screenshot 2018-11-13 at 12 14 37

However before this gets approved and merged, why was that overflow rule added? Usually it's a self-clearing thing, but wanted to make sure why so we don't cause other regressions.

### DIFF
--- a/style-editor.css
+++ b/style-editor.css
@@ -29,11 +29,6 @@ body .wp-block[data-align="full"] {
   body {
     padding-top: 0;
   }
-  body :not(.editor-inner-blocks) > .editor-block-list__layout,
-  body .editor-post-title {
-    padding-left: 0;
-    padding-right: 0;
-  }
   body .editor-writing-flow {
     padding-top: 50px;
   }
@@ -44,11 +39,6 @@ body .wp-block[data-align="full"] {
 }
 
 @media only screen and (min-width: 768px) {
-  body :not(.editor-inner-blocks) > .editor-block-list__layout,
-  body .editor-post-title {
-    padding-left: 46px;
-    padding-right: 46px;
-  }
   body .editor-block-list__layout,
   body .editor-post-title {
     max-width: 80%;
@@ -67,7 +57,7 @@ body .wp-block[data-align="full"] {
     position: relative;
     left: calc( -12.5% - 14px);
     width: calc( 125% + 116px);
-    max-width: calc( 125% + 116px);
+    max-width: calc( 125% + 115px);
   }
   body .wp-block[data-align="right"] {
     max-width: 125%;
--- a/style-editor.css
+++ b/style-editor.css
@@ -27,8 +27,8 @@ body .wp-block[data-align="full"] {
 
 @media only screen and (min-width: 600px) {
   body .wp-block[data-align="full"] {
-    position: relative;
-    left: 45px;
+    width: calc( 100% + 90px);
+    max-width: calc( 100% + 90px);
   }
 }
 
--- a/style-editor.css
+++ b/style-editor.css
@@ -26,12 +26,6 @@ body .wp-block[data-align="full"] {
 }
 
 @media only screen and (min-width: 600px) {
-  body {
-    padding-top: 0;
-  }
-  body .editor-writing-flow {
-    padding-top: 50px;
-  }
   body .wp-block[data-align="full"] {
     position: relative;
     left: 45px;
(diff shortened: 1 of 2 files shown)


kjellr:
Blargh. Thanks for catching that. I didn't test Unified Toolbar mode when I made the change. 

That was added yesterday in #575, in an effort to solve #564. The issue with removing it entirely is that there is some overflow that appears at various breakpoints: 

Between `600-768px`:

small-screens

And also at some seemingly-random larger sizes too, but only for alignfull blocks:

large-screens

I did some quick digging for the root of the problem yesterday, but nothing obvious jumped out. The alignfull one is likely something to do with my math (😄), but I'm not totally sure which margins/padding are causing the overflow for the smaller breakpoints. It could definitely use another set of eyes if you have a little time. 

In any case, if we can remove this overflow declaration entirely, I suggest we revert #575, as the margin solution there (applying the margin directly to the `.editor-styles-wrapper`) was more clean.

jasmussen:
Ah yes, I've struggled with that issue as well, and it's a tricky one. I can look at that issue separately for sure. 

> In any case, if we can remove this overflow declaration entirely, I suggest we revert #575, as the margin solution there (applying the margin directly to the .editor-styles-wrapper) was more clean.

Did you write in the wrong ticket name there?

kjellr:
> > In any case, if we can remove this overflow declaration entirely, I suggest we revert #575, as the margin solution there (applying the margin directly to the .editor-styles-wrapper) was more clean.
> 
> Did you write in the wrong ticket name there?

Whoops. Yes, I meant #565. 😄

jasmussen:
Okay, thank you for the context! 

In https://github.com/WordPress/twentynineteen/pull/575/commits/d78c86d3cc7037b7fd0609f3f88377cbd200399b, I tweaked things a bit. 

First off, I believe I fixed the issue with fullwide images by subtracting 1px from the fullwide max-width math. What was going on was that the combination of calc and percentage values caused the rounding engine to _round up_ on odd viewport widths. The downside here is that it's not beautifully full-width all the time, now it's 1px off on even viewport widths. But honestly I doubt few will notice, and I think it's a good tradeoff to make.

Similarly, the 46px padding left and right of blocks was reset in the editor style. I understand why it was, but due to the presence of side UI (movers and drag handle to the left), those paddings have to be there. By removing the override rules in the editor styles, this fixed it. 

I'll be taking a look at some additional improvements here. I think the balance we need to find is between that of providing an editor style that is as close to 1:1 with the frontend as possible, while still not writing _new software_ to get there (via overrides).

kjellr:
Cool, this is looking good!

> First off, I believe I fixed the issue with fullwide images by subtracting 1px from the fullwide max-width math. What was going on was that the combination of calc and percentage values caused the rounding engine to round up on odd viewport widths. The downside here is that it's not beautifully full-width all the time, now it's 1px off on even viewport widths. But honestly I doubt few will notice, and I think it's a good tradeoff to make.

That sounds fine by me, but I'm going to ping you when we get the but report about it. 😄 

> Similarly, the 46px padding left and right of blocks was reset in the editor style. I understand why it was, but due to the presence of side UI (movers and drag handle to the left), those paddings have to be there. By removing the override rules in the editor styles, this fixed it.

Ok — that's also a fine tradeoff. I'd rather things worked as expected than have them be an exact 1:1 match between the front and back end at this point. And it's only an issue at those middle 600-768px breakpoints.

jasmussen:
> I'm going to ping you when we get the but report about it. 😄

I would expect nothing less!

> Ok — that's also a fine tradeoff. I'd rather things worked as expected than have them be an exact 1:1 match between the front and back end at this point. And it's only an issue at those middle 600-768px breakpoints.

Yes, already now I think TwentyNineteen is _beyond_ what is necesary in an editor style. Which I mean as a great compliment — but we should also be careful to not try and do too much to get there, lest we end up having to redo things when Gutenberg changes upstream.

kjellr:
I pushed a couple small fixes: 

- Moved the wrapper margins back to where they were before #595 was merged. This is cleaner, and means we don't have to override the top padding of `.edit-post-visual-editor`.
- Fixed the width of `alignfull` blocks on breakpoints between `600-768px`.

kjellr:
Awesome. This seems pretty clean, so I'll merge in for now and we can improve as we go. Thanks @jasmussen!

PR #577, opened 2018-11-13
Fix search input style on iOS Safari

youthkee:
Style and look-and-feel of search input is defferent with other input elements on iOS Safari. This PR fixes it.

## Before

twentynineteen-search-input-style1

## After

twentynineteen-search-input-style2

### DIFF
--- a/sass/forms/_fields.scss
+++ b/sass/forms/_fields.scss
@@ -20,17 +20,16 @@ textarea {
 	box-sizing: border-box;
 	outline: none;
 	padding: #{.36 * $size__spacing-unit} #{.66 * $size__spacing-unit};
+	-webkit-appearance: none;
+	outline-offset: 0;
+	border-radius: 0;
 
 	&:focus {
 		border-color: $color__link;
 		outline: thin solid rgba( $color__link, 0.15 );
 		outline-offset: -4px;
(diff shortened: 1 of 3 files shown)


sass/forms/_fields.scss
```diff
@@ -28,6 +28,14 @@ textarea {
 	}
 }
 
+input[type="search"] {
+	-webkit-appearance: none;
```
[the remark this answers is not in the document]
youthkee: Thanks for your suggestion!
I added a few changes for that.

style-rtl.css
```diff
@@ -964,6 +967,25 @@ textarea:focus {
   outline-offset: -4px;
 }
 
+input[type="text"]::-webkit-search-decoration,
```
kjellr: One more small change: This `&::-webkit-search-decoration` property can be separated out so its specific to `input[type="search"]`. Since that property doesn't apply to non-search input fields, most of these auto-compiled styles aren't necessary. 

Thanks again!

youthkee: Sorry for forgetting that point.
I fixed it again!

kjellr: Looks great, thank you @youthkee!

PR #585, opened 2018-11-13
Try: Use non-breaking space override globally

kjellr:
A followup to #463.

This PR takes the `&nbsp;` font override from our editor styles and applies it to the front end as well. It aligns the width of Hoefler Text non-breaking spaces in Chrome with their width in other browsers, and if all goes well, it should have no effect on the front end except when viewing a site in Chrome on a Mac. 

In that case, you'll notice that `&nbsp` characters are shorter. For instance, nested submenu items will appear differently: 

**Before**
screen shot 2018-11-13 at 3 54 56 pm

**After**
screen shot 2018-11-13 at 3 55 09 pm


kjellr:
Thanks!

PR #589, opened 2018-11-14
Fix priority+ menu in customizer with a full customizer refresh

allancole:
Addresses #558 and improves on the solution in #568 

I found a way to get vanilla javascript to listen for jQuery customizer events :-) That lead me to this fix which forces the customizer to fully reload when making partial edits to the menu. This is necessary to re-run the priority+ function only _after_ the menu content has fully been fully updated in the customizer. 

Ideally, we would simply re-run the priority+ function on partial refreshes in the customizer, which would create a faster experience. However, partial refreshes only update the main navigation list wrapper. Any items that were moved to the hidden list by the priority+ function, get stuck in the dom on partial refresh leaving an extra, orphaned (...) menu link. This fix prevents that from happening with the downside of a slower customizer experience when changing the menu.


js/priority-menu.js
```diff
@@ -35,94 +72,121 @@
...
+		debounce( function() {
+			if ( isResizing ) {
+				return;
+			}
 
-	toggleButton.addEventListener('click', function() {
-		toggleElementVisibility(hiddenList);
-	});
+			isResizing = true;
+			setTimeout( function() {
```
youknowriad: Why do we need setTimeout here? debounce and setTimeout are similiar functions, you can pass `150` as a second argument of debounce and I think It will have the same effect

allancole: Ahhh, good catch—no need for setTimeout. I’ll sort that out 👍

kjellr:
This doesn't seem to be working for me: I'm still seeing the double icons when updating the menu: 

customizer

customizer-alt

PR #591, opened 2018-11-14
Add front end padding to column blocks.

### LINKED ISSUE #517: No left and right margin for fully aligned columns
We should add some margin (left and right) for fully aligned columns. It is fine not having margin for images, but text columns should have margins.

[Example](https://user-images.githubusercontent.com/5901923/48051046-3aab6f80-e1a4-11e8-8a58-826e57c10b9d.png)

Something like

```
.wp-block-columns {
		
	&.alignfull {
		margin-left: $size__spacing-unit;
		margin-right: $size__spacing-unit;
	}
}
```

in blocks.scss.

Only for the columns wrapper.

Or we might target first and last childs column blocks...

kjellr:
Fixes #517.

This PR seeks to add some padding to our `alignfull` columns so that they match up more closely with the editor.

**Appearance in editor:**

_Note: this is based on the columns appearance seen in the #575 branch. That happens to correct a bug that inserts too-wide margins on the columns block. As a result, it might be helpful to wait until that PR is merged to text/merge this one._

screen shot 2018-11-14 at 11 02 10 am

**Current appearance in front end:**

screen shot 2018-11-14 at 11 08 08 am

**Updated appearance in front end:** 

_Note: This does not line up exactly. But that's mostly because the editor seems to add inconsistent margins on the left and right side of the columns block when it's full-screen. I've filed a Gutenberg bug for that (https://github.com/WordPress/gutenberg/issues/11869), but in the meantime we don't want to replicate that error._

screen shot 2018-11-14 at 11 02 23 am

---

This also adds padding to smaller screens: 

**Before:** 
screen shot 2018-11-14 at 11 30 43 am

**After**
screen shot 2018-11-14 at 11 31 06 am

### DIFF
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -681,6 +681,11 @@
 	//! Columns
 	.wp-block-columns {
 
+		&.alignfull {
+			padding-left: $size__spacing-unit;
+			padding-right: $size__spacing-unit;
+		}
+
 		@include media(tablet) {
 			.wp-block-column > * {
 
@@ -706,11 +711,6 @@
 				padding-left: calc(2 * #{$size__spacing-unit});
 				padding-right: calc(2 * #{$size__spacing-unit});
 			}
-		}
-
-		&.alignfull, {
-			padding-left: $size__spacing-unit;
-			padding-right: $size__spacing-unit;
 		}
 	}
 
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -700,6 +700,17 @@
 					margin-right: 0;
 				}
 			}
+
+			&.alignfull,
+			&.alignfull .wp-block-column {
+				padding-left: calc(2 * #{$size__spacing-unit});
+				padding-right: calc(2 * #{$size__spacing-unit});
+			}
+		}
+
+		&.alignfull, {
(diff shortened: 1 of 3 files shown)


jasmussen:
Worth noting as mentioned also in the upstream PR, the additional left and right padding that's present  in the editor is intentional, even though it means the preview is not 1:1 correct: https://github.com/WordPress/gutenberg/compare/improve/columns?expand=1#diff-49ac41bc72e24e343a451fe254801c0fR16

kjellr:
Interesting — I've weighed in on the upstream PR, and will hold off until we see what happens there.

kjellr:
I gave this a re-test, and I think it's still a good improvement to include on our end. It only effects the front-end, and just adds a little padding so that text columns don't bump against the side of the viewport. It also happens to make the front-end look a little like the editor, but that's mostly a side effect. 

**Before:**
screen shot 2018-11-16 at 10 19 18 am 
screen shot 2018-11-16 at 10 19 34 am

**After:**
screen shot 2018-11-16 at 10 18 45 am
screen shot 2018-11-16 at 10 19 03 am


(@jasmussen looking at your front-end screenshots above, I  wonder if there was a caching issue or something  —  I  don't see this PR's changes reflected in them).

kjellr:
🎉 Thanks!

PR #598, opened 2018-11-14
checkbox for color filter

### LINKED ISSUE #570: Black filter missing on blog page
**Summary:**
Applying the black filter to featured images results in only featured images on the individual post page from using this filter. When viewing these featured images from the main blog page they have no filter.

**Steps to reproduce:**
1. Go to Customize > Colors
2. Under Color Scheme choose _Custom_
3. Apply the **Black Filter** to your featured image

**Expected behavior:**
I expect the black filter should be applied to featured images on the blog roll and individual post pages. I expect this because if I use the colored filter option then it is applied at both of these location.

**Actual behaviour:**
The black filter only applies to the featured image when viewed from the individual post page. The filter does not get applied to the featured image on the blog roll.

**Screenshots:**

Colored filter -
 * Blog roll: 
screen shot 2018-11-12 at 7 35 46 pm

 * Individual post:
screen shot 2018-11-12 at 7 35 08 pm

--

Black filter -
 * Blog roll: 
screen shot 2018-11-12 at 7 36 00 pm

 * Individual post:
screen shot 2018-11-12 at 7 23 34 pm

LittleBigThing:
Fixes #570 

Use a checkbox instead of a radio button control for the featured image color filter in the Customizer


kjellr:
Awesome — this is looking great: 

screen shot 2018-11-14 at 3 54 15 pm

Any chance we can add a "Featured Image Color Filter" title above the checkbox like the one in the comp?

screen shot 2018-11-14 at 3 56 34 pm

LittleBigThing:
[Just commented](https://github.com/WordPress/twentynineteen/issues/570#issuecomment-438815572) about the title. :-)

allancole:
I like it better with the title but I’m fine with it not being there if thats just how the API is designed to deal with checkboxes.

LittleBigThing:
This is how it looks when using it the standard way with label and description:

screenshot 2018-11-14 at 22 41 45

kjellr:
Thanks @LittleBigThing — let's stick with the way it is in the PR in that case. I think that's better than changing the label to "Featured Image Color Filter": The label in the PR is phrased as an action, which lines up with other common checkbox options ("Display Site Title and Tagline", for instance).  

One minor layout thought: Is it possible to increase the margin between the checkbox and the hue picker? Right now, there's a comfortable amount of space in between controls when the Default color is active, but when the hue picker is open the checkbox is right up next to it.

artboard

allancole:
I’m starting to think we should consolidate this even more so that it feels like one single option called “Primary Color” instead of “Color Scheme.” This way we don’t need a second title for the Custom Color and the spacing for the filter checkbox is no longer necessary since its all part of the same option

primary-color-dialogue

cc: @kjellr @LittleBigThing

LittleBigThing:
Yes, that’s it! Great idea @allancole.

If we are not going to add another color scheme (and I suppose we won’t), then this would probably be the simplest UI. 👍🙂

allancole:
Indeed! Mind revising this PR @LittleBigThing? Or I can take care of it if you like.

LittleBigThing:
I’m going to update the PR with these suggestions.

LittleBigThing:
Working on the PR and a bit in doubt about changing the name of the theme mods?
- `primary_color` instead of `color_scheme`
- `primary_hue` instead of `colorscheme_primary_hue`

This is of course changing all instances of the theme mods.

allancole:
Ahh, good catch. I think it’s fine to change the theme mod naming convention since we’re still _technically_ in Beta.

LittleBigThing:
Yey! Let's hope we're not breaking the theme in RC1. ;-)

Issue #567, opened 2018-11-12
Visited links with a button class have the same value for color and background properties


wvega:
I noticed that the text from any visited link with the class `button` can no longer be read because the links are styled using the same value for `color` and `background-color` properties:

<img width="746" alt="screen shot 2018-11-12 at 3 21 28 pm" src="https://user-images.githubusercontent.com/45068/48373121-d3ebf180-e68e-11e8-984d-39e76d82b463.png">

**Is this a use case that should be covered by the theme's stylesheet? or should I avoid using the `button` class on links?**

---

Twenty Seventeen, Twenty Sixteen, Twenty Fifteen, Twenty Thirteen, Twenty Twelve, Twenty Eleven and Twenty Ten don't seem to provide CSS rules for the `button` class. The links are displayed as regular links:

**Twenty Seventeen**
<img width="548" alt="screen shot 2018-11-12 at 3 24 52 pm" src="https://user-images.githubusercontent.com/45068/48373214-2a593000-e68f-11e8-8a05-badef8935dda.png">

However, Twenty Nineteen, Twenty Fourteen and several custom themes include rules to display those links as buttons:

**Twenty Fourteen**
<img width="502" alt="screen shot 2018-11-12 at 3 27 26 pm" src="https://user-images.githubusercontent.com/45068/48373601-47423300-e690-11e8-860c-713ac48df5d9.png">

**i-Spirit**
<img width="670" alt="screen shot 2018-11-12 at 3 31 17 pm" src="https://user-images.githubusercontent.com/45068/48373596-414c5200-e690-11e8-8cc7-b902b4467a61.png">

**OceanWP**
<img width="627" alt="screen shot 2018-11-12 at 3 35 39 pm" src="https://user-images.githubusercontent.com/45068/48373711-a56f1600-e690-11e8-91f3-9897253652e4.png">

kjellr:
Should be all set now that #600 has been merged. Thanks!

PR #600, opened 2018-11-14
Adds button styles to legacy wp `.button` class elements

allancole:
Adds a fix for #567 

Many WP themes include a `a.button` class to allow users to make normal links look like buttons. This PR fixes an issue where `a.button:visited` button links would run into a color conflict. 

Before:
image

After: 
image

This PR fixes the color conflict but it does not include styles for the editor since the `a.button` markup is the same as normal text-link markup in the editor. It does not use the same `editable` element wrappers that are used in the Button Block, and so there’s a simple way to make the `a.button` styles match the Button Block styles while still being editable text. The end result is this: 

fix-visited-button-links-2

@kjellr or @jasmussen how do you think we should best deal with this?

### DIFF
--- a/inc/color-patterns.php
+++ b/inc/color-patterns.php
@@ -82,7 +82,7 @@ function twentynineteen_custom_colors_css() {
 		 * - Widget links
 		 */
 		a,
-		a:visited,
+		a:not(.button):visited,
 		.main-navigation .main-menu > li,
 		.main-navigation ul.main-menu > li > a,
 		.post-navigation .post-title,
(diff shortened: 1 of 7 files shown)


jasmussen:
I don't personally have a strong opinion here, so I will defer to you and Kjell on the necessity of this.

However on principle, the Gutenberg button block is created exacly to make unnecessary that old behavior. So in that vein it doesn't feel like an issue that TwentyNineteen _has_ to address.

kjellr:
I agree with Joen there. I'm fine including this because it's essentially done already, but in general I wouldn't want us to spend additional time on this sort of thing, since it's already present (and better) in Gutenberg.

One small issue to note: I'm seeing the incorrect text color on hover...

hover

allancole:
Just fixed that issue ^^ @kjellr. 

fix-visited-button-links-3

PR #603, opened 2018-11-15
Remove unnecessary attributes from priority nav ellipsis

westonruter:
Introduced in #485 for the priority menu.


westonruter:
See also related issue: https://github.com/WordPress/twentynineteen/issues/604

PR #607, opened 2018-11-15
Submenu: Fix shift-tab navigation and animation

### LINKED ISSUE #266: Main navigation menu not fully operable with a keyboard

kjellr:
Fixes #266.

When reverse-tabbing through the menu currently, you end up skipping over the submenu items: 

current

As @westonruter [noted](
https://github.com/WordPress/twentynineteen/issues/266#issuecomment-433560794), this is because we're using `display: none` on the submenu items when they're not in use. I don't think this is actually necessary, since we're also using `opacity: 0` and  `position: -9999px`. 

Removing this rule makes reverse-tabbing behave as expected: 

improved

There's one bug that still needs to be sorted out though — you'll notice that when you shift focus from a submenu to the child of its predecessor, the original submenu remains active. That's because we currently only remove the `is-focused` class from top-level menu items if their previous/next top-level item is selected:

https://github.com/WordPress/twentynineteen/blob/a2e298368f96524ed0e6452cfe4020175c75b11f/js/touch-keyboard-navigation.js#L304-L312

@allancole, can we adjust that to also remove the `is-focused` class if the child of a  previous/next item is selected as well? 

---

This PR also moves the fade-in animation back to where it should be, so it triggers on hover.  👍

### DIFF
--- a/js/touch-keyboard-navigation.js
+++ b/js/touch-keyboard-navigation.js
@@ -285,20 +285,20 @@
 			if ( event.target.matches('.main-navigation > div > ul > li a') ) {
 
 				// Remove Focused elements in sibling div
-				var currentDiv        = getCurrentParent( event.target, 'div' );
+				var currentDiv        = getCurrentParent( event.target, 'div', '.main-navigation' );
 				var currentDivSibling = currentDiv.previousElementSibling === null ? currentDiv.nextElementSibling : currentDiv.previousElementSibling;
 				var focusedElement    = currentDivSibling.querySelector( '.is-focused' );
 				var focusedClass      = 'is-focused';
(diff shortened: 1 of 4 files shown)


allancole:
The commits above fix how the `.is-focused` class is added/removed when using tabbed browsing in both directions, regardless of wether the focus is in a “Main menu” sub-menu or the “More menu” sub-menu. Tested this in IE11 and it checks out 👍. Gonna go ahead and merge.

PR #617, opened 2018-11-16
Increase mobile logo size and fix site-title link color

allancole:
Increasing the size of the logo on mobile screens, to make it more closely match the general size of avatars in “native” mobile apps. I also noticed that the .site-title link color changes to the custom color when it should remain black. This PR fixes that as well.

Before:
image

After:
image


allancole:
Done 👍

PR #618, opened 2018-11-16
Revert .button class link color to fix cascading issue

allancole:
Discovered an color cascading issue that came about while fixing a bug in #600. This change keeps that fix intact while restoring the intended colors to site-title, post-title and entry-date links.

Before:
image
----------
image
----------
image
----------

After:
image
----------
image
----------
image

### DIFF
--- a/inc/color-patterns.php
+++ b/inc/color-patterns.php
@@ -85,7 +85,6 @@ function twentynineteen_custom_colors_css() {
 		 * - Widget links
 		 */
 		a,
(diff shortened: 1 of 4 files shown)


joyously:
Did this one affect the print style link color also?

kjellr:
> Did this one affect the print style link color also?

I don't think so — that actually seems to be an issue with the print stylesheet.

allancole:
This ^^ commit should fix that @kjellr 

image

Issue #609, opened 2018-11-15
print styles still have problems


joyously:
When I first tried the fix for the print styles, the links did not have the href attribute, but I was seeing the comment form. I redownloaded the theme zip and tried again, but now I see the href on the links and no comment form. I looked at `blame` on the print.scss file, but couldn't tell what might be the problem.

What I am seeing now does not look like the example in PR #566. What I see is links have the href attribute, visited links are shown in color, and I'm not sure why two links show in another color. The comment Reply link is visible also.

![print-preview- 2018-11-15 15-39](https://user-images.githubusercontent.com/7957680/48583830-c2545500-e8ed-11e8-944f-29176b90780a.gif)

kjellr:
@joyously It'd be helpful to have a clear list of the issue you're seeing. Here's what I gather from your notes above:  

1. Links have the href attribute 
_I think this is the behavior we want, right? My screenshot before didn't include them because they're configured not to print for local links (This has not changed in the previous PR)_

2. Visited links are shown in color and I'm not sure why two links show in another color.
_I don't feel strongly about this, but don't mind changing it to all one color._

3. The comment Reply link is visible also
_We can change this._

Are there other print-related issues that you're seeing?

joyously:
Sorry, I thought it was clear. 
The GIF is from a localhost installation, so my links should look like yours did, right? It is showing the href for local links.
I think your summary is accurate. I was just confused because my previous test showed the comment form and no href, then it reversed (with no additional changes that I could find).

kjellr:
Thanks for the clarification. I can open a PR to fix # 2 and 3. 

> I was just confused because my previous test showed the comment form and no href, then it reversed (with no additional changes that I could find).

I might be mistaken about that. Upon further testing, I can't get the links to show up for me at all, in any environment. In any case, that last update didn't make any change to those rules so I'm not sure what would've changed. 

It appears the URL-adding styles were borrowed from [here](https://www.jotform.com/blog/css-perfect-print-stylesheet-98272/). I'm tempted to just remove that bit, since it seems inconsistent and hasn't been a part of previous themes.

PR #620, opened 2018-11-16
Fix minor print styles issues

kjellr:
Takes care of everything in https://github.com/WordPress/twentynineteen/issues/609 except for the links issue (which still confuses me).

- Removes a brown color that some links were using
- Hides the main navigation, comment reply, and comment edit links
- Removes a subtle background color for color-filtered headers that wasn't supposed to be there anymore

Before:
screen shot 2018-11-16 at 2 40 15 pm
screen shot 2018-11-16 at 2 41 02 pm

After:
screen shot 2018-11-16 at 2 32 29 pm
screen shot 2018-11-16 at 2 41 24 pm

### DIFF
--- a/print.css
+++ b/print.css
@@ -109,6 +109,11 @@ Andreas Hecht in https://www.jotform.com/blog/css-perfect-print-stylesheet-98272
   .comment .comment-metadata .edit-link {
     display: none;
   }
+  .entry .entry-content .wp-block-button .wp-block-button__link,
+  .entry .entry-content .button {
+    color: #000;
+    background: none;
+  }
   /* Site Header (With Featured Image) */
   .site-header.featured-image {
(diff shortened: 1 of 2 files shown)


kjellr:
> This looks great but just noticed that buttons should probably get a print treatment. I’d be okay if they looked just like normal sans-serif links, but defer to your judgement :-)

Alright, this should do it. I can't have them inherit the normal link color because we're not defining it in the stylesheet, and there are too many cascading styles to override. So instead, I've just made them all black and removed the background. If someone has set a custom link color in Gutenberg, that's used instead though. 

Front end: 
screen shot 2018-11-16 at 3 46 23 pm

Print: 
screen shot 2018-11-16 at 3 46 05 pm

PR #383, opened 2018-10-29
Add non-latin language support

nielslange:
Fixes #29 

Add non-latin language support for:

- Arabic —> ar: العربية, ary: العربية المغربية, azb: گؤنئی آذربایجان, ckb: كوردی‎, fa-IR: فارسی, haz: هزاره گی, ps: پښتو
- Chinese —> zh_HK: 香港中文版, zh_TW: 繁體中文, zh_CN: 简体中文
- Cyrillic —> be: Беларуская мова, bg_BG: Български, kk: Қазақ тілі, mk_MK: Македонски јазик, mn: Монгол, ru-RU: Русский, sah: Сахалыы, sr_RS: Српски језик, tt_RU: Татар теле, uk: Українська
- Devanagari —> bn_BD: বাংলা, hi_IN: हिन्दी, mr: मराठी, ne_NP: नेपाली
- Greek —> el: Ελληνικά
- Gujarati —> gu: ગુજરાતી 
- Hebrew —> he_IL: עִבְרִית
- Japanese —> js: 日本語
- Korean —> ko_KR: 한국어
- Thai —> th: ไทย
- Vietnamese —> vi: Tiếng Việt

PR #621, opened 2018-11-16
Fix non-latin language fonts effecting the editor UI

### LINKED ISSUE #602: Implement a better font-family solution for non-latin languages
I think that 635a5bb of #383 is very rough.
In the non-latin languages, the situation that all `font-family` are unified into only one is unbearable.
Especially `<code>` and `<pre>` should be `monospace` even in non-latin languages.

I propose more fundamental solution.  The size of style.css will be larger than now, but I believe that maintenance is not so difficult.

A patch is attached.
This patch is merely to show the direction of the solution and is not complete.
Therefore, only the part related to $font__body and only about a few languages are described.

https://gist.github.com/mako09/bf919903b7e4476279e77d2d0c2d9099

allancole:
Revises #383 and attempts to fix #602 so that the non-latin language fonts effect the Gutenberg editor less aggressively. 

Before (notice the broken icons):
image

After: 
image

This change also limits the new fonts to editable text areas, as opposed to changing the fonts on the entire block wrapper which can broke parts of the editor UI. In both screenshots, you’ll notice where the Japanese font stack is being added. In the After screenshot, its added to the `.mce-content-body` class so that these language-specific fonts don’t interfere with the Gutenberg UI. 

More info here: https://github.com/WordPress/twentynineteen/issues/602#issuecomment-439407971

@mako09 could you review this and make sure it works for you?

### DIFF
--- a/sass/mixins/_mixins-master.scss
+++ b/sass/mixins/_mixins-master.scss
@@ -89,88 +89,88 @@
 
 /* Fallback for non-latin fonts */
 
-@mixin non-latin-fonts( $wrapper_classname: '.site' ) {
-	
+@mixin non-latin-fonts( $wrapper_selector: '.site *' ) {
+
 	/* Arabic */
-	html[lang="ar"] #{$wrapper_classname} *,
-	html[lang="ary"] #{$wrapper_classname} *,
-	html[lang="azb"] #{$wrapper_classname} *,
-	html[lang="ckb"] #{$wrapper_classname} *,
-	html[lang="fa-IR"] #{$wrapper_classname} *,
-	html[lang="haz"] #{$wrapper_classname} *,
-	html[lang="ps"] #{$wrapper_classname} * {
+	html[lang="ar"] #{$wrapper_selector},
+	html[lang="ary"] #{$wrapper_selector},
+	html[lang="azb"] #{$wrapper_selector},
+	html[lang="ckb"] #{$wrapper_selector},
+	html[lang="fa-IR"] #{$wrapper_selector},
+	html[lang="haz"] #{$wrapper_selector},
+	html[lang="ps"] #{$wrapper_selector} {
 	  font-family: Tahoma, Arial, sans-serif !important;
 	}
 
 	/* Cyrillic */
-	html[lang="be"] #{$wrapper_classname} *,
-	html[lang="bg-BG"] #{$wrapper_classname} *,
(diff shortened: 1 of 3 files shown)


sass/mixins/_mixins-master.scss
```diff
@@ -89,90 +89,97 @@
...
 	/* Vietnamese */
-	html[lang="vi"] #{$wrapper_classname} * {
+	html[lang="vi"] #{$wrapper_selector} {
 	  font-family: 'Libre Franklin', sans-serif !important;
 	}
+
+	/* Use standard monospace font for all non-english code and pre */
+	html:not([lang^="en-"]) #{$wrapper_selector} code,
+	html:not([lang^="en-"]) #{$wrapper_selector} pre {
+		font-family: monospace;
```
kjellr: I think this needs an `!important` rule in order to override the `!important` stuff above.

style-editor-customizer.scss
```diff
@@ -10,4 +10,4 @@ NOTE: This file is automatically populated with additional styles if the user se
 
 /** === Non-Latin font fallbacks === */
 
-@include non-latin-fonts( '.wp-block' );
\ No newline at end of file
+@include non-latin-fonts( '.wp-block .mce-content-body' );
```
kjellr: This isn't quite working when I test — blocks aren't inheriting the correct styles until I click on them. 

I think blocks don't actually get the `.mce-content-body` class until they've been active once. You may need to also specify `.editor-rich-text__tinymce` to get that working. You'll also need to target `editor-post-title__input`, since that doesn't get that class either. 😕

kjellr:
@allancole —  I pushed a couple commits to fix those two issues I noted last night. It mostly works well in the front end and in Gutenberg, but this approach doesn't appear to work at all in the Classic editor right now. 😞 

In any case, it seems a little fragile in general — I'm wondering if we need to think through a  different solution, as @mako09 suggests.

allancole:
Yeah, it seems this might have to happen another way, for Gutenberg. 

Also, it could be worthwhile to research how well Gutenberg deals with non-latin fonts without the theme. If it has it’s own built-in solution for non-latin fonts, we can probably just borrow that and tweak it to use our fonts. But if it doesn’t have a built-in solution, anything we do from the theme will likely fall out of sync if/when Gutenberg rolls out its own solution later. 

With that in mind, we might only need to address this issue for the frontend at this time.

PR #623, opened 2018-11-16
Move the `&nbsp;` fix to editor-only styles

allancole:
Since the `NonBreakingSpaceOverride` is not needed on the frontend, we shouldn’t load it at all unless we’re in the editor. This PR makes it so that fix is only added to the Gutenberg editor, which will make our `style.css` file a good bit lighter on the front end. It also removes a duplicate `colors.scss` include.


kjellr:
> Since the NonBreakingSpaceOverride is not needed on the frontend, we shouldn’t load it at all unless we’re in the editor. 

@allancole — It's possible you didn't see the PR, but @jasmussen and I merged that into the front end in https://github.com/WordPress/twentynineteen/pull/585 the other day. While not _definitely_ required, it  does fix some little quirks like the menu spacing noted in the PR. Since the custom font only covers one character, it should add just a couple kilobytes to the site's load. I'd prefer to keep it in, but not a huge deal either way.

PR #633, opened 2018-11-18
Implement a better font-family solution for non-latin languages

### LINKED ISSUE #602: Implement a better font-family solution for non-latin languages

mako09:
Fixes #602

* Obsoletes #383 #621
* I am not sure what is a suitable Serif font for each language. These definitions in sass/variables-site/_fonts.scss should be updated according to native users' opinion.


kjellr:
Thanks for bringing this into a PR, @mako09, and for your dedication to getting this right. 

I have a few suggestions and comments: 

- I think there may be a syntax issue somewhere: Some of the font family values render with quotation marks around the whole set of values. For instance: `"'Sukhumvit Set', 'Helvetica Neue', helvetica, arial, sans-serif"` instead of `'Sukhumvit Set', 'Helvetica Neue', helvetica, arial, sans-serif`. This seems to be preventing some of the sans-serif fonts from loading.  
- I see you're specifying a number of Noto fonts. Since we aren't specifically loading those at all on the frontend, most people won't end up seeing those fonts. I'm not sure it makes sense to use those here. 
- If we don't already know an appropriate substitute font, I'd just use the font stack we were using before for each language option. Even if it means using some sans serif fonts in place of serif fonts for now.
- Speaking of that, in general I don't mind using sans-serif fonts everywhere for some languages. For example, we've already [heard feedback](
https://github.com/WordPress/twentynineteen/issues/29#issue-370676636) that it may not be appropriate to use a serif font for Japanese body text.

While this is in many ways a cleaner implementation, my main concern is just timing: RC1 is shipping tomorrow, so if we decide to move forward with this method, it probably makes sense to need to hold off and ship it after that. 

cc @allancole for additional thoughts.

mako09:
@kjellr Thank you for your suggestions.

> * I think there may be a syntax issue somewhere: ...
> * ... I think that may be unrelated to the syntax issue above.

I fixed them.

> * I see you're specifying a number of Noto fonts. ...
> * If we don't already know an appropriate substitute font, ...
> * Speaking of that, in general I don't mind using sans-serif ...

Yes, I don't know appropriate fonts, Therefore 'Noto *' are just placeholders.
I hope they will be decided with further opinion.

kjellr:
Hi @mako09! As noted in related threads, we're migrating this work over to Trac. I've opened this issue with the changes from #728, which include much of the general approach in this PR. 

https://core.trac.wordpress.org/ticket/45731

I'm going to close this PR, and we can continue tracking updates in that thread. Thanks again!

Issue #630, opened 2018-11-17
Limit width of `full` images to original image width


mor10:
When an image is set to `full` alignment, limit the displayed with of the image to the original width. Right now the image continues to match `100vw` even when the image file itself is not wide enough causing artifacts and a "blurry" look.

kjellr:
> When an image is set to full alignment, limit the displayed with of the image to the original width. 

I have a couple reasons to be hesitant about making this change:

1. Gutenberg already sets some precedent by not doing this in the editor. All themes that use Gutenberg will need to think about this. If we implement a solution, we'll need to apply it to both the front and back end. 
2. From a user expectations standpoint, if someone specifies that they want a "full width" image, I think they should get a full width image. In this case, full width images will stretch the full width in the editor, they'll see immediately that the image will be blurry/pixellated (unlike #208). If they're not okay with this effect, they can choose a different alignment option. 

I'll cc @allancole for a  final opinion, but I lean towards marking this as `wontfix`.

joyously:
Does that take into account that the user could have any window size while editing? I know a lot of people that don't know to consider that what they see as they create a page could be very different for a visitor on a different device. The editor is very skinny in my usual window size, so a full width image looks fine. But when a visitor has a huge desktop monitor with fullscreen browser windows, what does he see? 
My opinion: the editor should show a notice that the image is xxx wide and some viewers will see a stretched version. Then the informed user can decide to continue that way or not.

mor10:
@kjellr this is a problem with the concept of "full width" as presented by Gutenberg. Displaying an image as wider than the image will be perceived by the end-user as a broken user experience. Expected behavior is for the image to be displayed accurately. Stretching it beyond its intrinsic dimensions goes against how images should behave. In my opinion.

allancole:
This is a pretty tricky one and while it’s not ideal for images to display with blurriness, a decision on this comes down to who is ultimately responsible for handling low quality images. Regardless of any expected behaviors presented by Gutenberg (or the theme), the source of this problem is the low-quality image itself. Any full-width scaling we do or _don’t_ do will not correct that original image size. Whether we leave it as-is with a stretched image, or limit the image to its original width, the experience is broken because the image is simply too small for full-width. That makes this decision much more about how to inform a user that their image is too low-quality for `full` alignment. 

[...]

PR #638, opened 2018-11-18
Limit width of `wide` and `full` images to original image width

mor10:
Addresses #630.

Limit displayed with of `alignwide` and `alignfull` images to the original width of the uploaded image file to avoid stretching of images.


mor10:
The bundled block stylesheet specifies the width. This overrides that style rule. Test the PR and you'll see.

mor10:
Closing as #630 is closed.

PR #639, opened 2018-11-19
Fix site logo selective refresh issue

allancole:
Found a bug where the menu would break when updating the site-logo. This was happening because the conditional for re-running the priority+ function on selective refresh was happening on _every_ partial refresh. This fix makes the conditional more specific so that it only re-runs the priority+ function when the menu is refreshed. 

Before:
site-logo-menu-error

After: 
site-logo-menu-fixed

### DIFF
--- a/js/priority-menu.js
+++ b/js/priority-menu.js
@@ -147,15 +147,21 @@
 		var hasSelectiveRefresh = (
 			'undefined' !== typeof wp &&
 			wp.customize &&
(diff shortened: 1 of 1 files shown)


kjellr:
I'm having trouble reproducing this one in `master`. When I remove the site logo and make other changes, I don't get console errors like the ones you're seeing. Can you clarify the steps needed to recreate the issue?  

Thanks!

Issue #111, opened 2018-10-17
Handle text overflow on <pre> tags


audrasjb:
Texts inserted via the `code` block are not visible if they overflow their container.
Please add overflow-x property to `<pre>` tag CSS rules :)

nielslange:
Looks like a duplicate of #84

PR #642, opened 2018-11-19
Add additional colors to our editor swatches

kjellr:
As per  @jasmussen's suggestion here: 
https://github.com/WordPress/gutenberg/issues/12024#issuecomment-439877180y

This PR adds: 

- Black
- Dark Gray (`#111`)
- Light Gray (`#767676`)
- White

Note: Dark Gray and Black are very similar and I doubt many users I'll be able to tell the difference without hovering. But it felt weird to not include black if we're including white, and it felt weird not to include Dark gray since it's used for just about all text on the site. I'm interested to hear other thoughts on the matter. 🙂

**Before:** 
screen shot 2018-11-19 at 8 16 04 am

**After:**
screen shot 2018-11-19 at 8 14 37 am


kjellr:
> The more the better though, and I think we should add at least one maybe two more non gray colors. In that pr above, I added a lighter blue version of the spot color, which uses the same hue and saturation. Isn't that a good way to do it?

That is, but I think what I'd do here is use the darker hover color we're auto-generating instead, since we're already generating and using that in the theme. Will push a few updates shortly...

kjellr:
Added another custom color, and made it so that these colors appear correctly on the front end: 

screen shot 2018-11-19 at 9 04 59 am

Pullquote blocks are currently _not_ working right. Unlike some other blocks that allow for custom colors, they don't get a `has-background` color class, so I may have to use a lot of extra selectors to make sure our styles for each of the new custom colors don't get overridden by mistake. Digging in now...

jasmussen:
Dig it, totally shippable. I'd love to see more colors that are not blue or gray, like one additional spot color, but it's not critical and this is good to go! Thanks Kjell! 

Did you have any thoughts on adding a border to the regular style pullquote when a main color is set?

kjellr:
>  Did you have any thoughts on adding a border to the regular style pullquote when a main color is set?

Not yet. To start, I'd say we should start by just enabling the default borders when someone selects a main color. Would you mind spinning up a PR to get that working by any chance? If not, I can get it going once I finish up this PR.

allancole:
I don’t think it makes sense to add both `black` and `#111`. They’re just so close that we might as well just choose one. I say we use `#111` and call it “Black” in the UI since the theme already uses `#111` as if it were Black. Users could still select perfect black with the color-picker if they wanted to.

I think we should remove the darkened hover color as an available option as well—it’s effectively redundant against the primary color, imo. That leaves us with just: 
- Blue (Primary color based on customizer settings)
- Black (`#111111`)
- Grey (`#767676`)
- White (`#ffffff`)

> I'd love to see more colors that are not blue or gray, like one additional spot color, but it's not critical and this is good to go! 

I think it might be more useful to bring in a few arbitrary colors here too. If we want users to get a good feel for the Gutenberg experience, it makes more since to bring in all or some of the colors from the default Gutenberg palette so that there’s even less confusion about radio buttons as [mentioned on the Gutenberg comment](https://github.com/WordPress/gutenberg/issues/12024#issuecomment-439877180y): 

image

In the interest of time, maybe we could just go with 6 total colors + the rainbow (so its on it’s own line)? The 4 colors mentioned above and two arbitrary ones from the Gutenberg default.

allancole:
Also seeing some inconsistencies in the solid-color Pull quote. Looks like the selectors are still being overwritten:

**Backend**
image

**Frontend:**
image

kjellr:
> I don’t think it makes sense to add both black and #111. They’re just so close that we might as well just choose one. I say we use #111 and call it “Black” in the UI since the theme already uses #111 as if it were Black. Users could still select perfect black with the color-picker if they wanted to.

Calling `#111` black seems a little disingenuous to me. I'm okay cutting `#000`out, but keeping `#111` as "Dark Gray".

> In the interest of time, maybe we could just go with 6 total colors + the rainbow (so its on it’s own line)? The 4 colors mentioned above and two arbitrary ones from the Gutenberg default.

Because we allow people to choose just one custom site color, it'll seem weird to have a couple non-editable random colors highlighted in there all the time.

I think the simplest, most conservative way for us to get this in for RC1 would be to include just the four you mentioned above: 

- Blue (Primary color based on customizer settings)
- Dark Gray (`#111111`)
- Light Gray (`#767676`)
- White (`#ffffff`)

Those will reflect the front end colors, which makes sense. Then we can revise later if needed. Does that sound good?

allancole:
Separate from the actual colors, Pullquotes are breaking because Gutenberg is adding ALL the color classes to the frontend instead of just the one that’s actually selected by the user. Take this block in the editor for example:

image

Expected markup on the front end:
`<figure class="wp-block-pullquote is-style-solid-color has-light-gray-background-color">`

Actual markup:
`<figure class="wp-block-pullquote has-secondary-background-color has-dark-gray-background-color has-light-gray-background-color has-primary-background-color is-style-solid-color">`

If that isn’t fixed in Gutenberg, custom-colors will always override the custom settings because `.has-primary-background-color` is always there: 

image
(see the overrides in the inspector)

kjellr:
> Separate from the actual colors, Pullquotes are breaking because Gutenberg is adding ALL the color classes to the frontend instead of just the one that’s actually selected by the user. 

Yikes, that doesn't seem good. It sounds like that's the cause of [the issue you noted before](
https://github.com/WordPress/twentynineteen/pull/642#issuecomment-439971536), so I think the  rest of the PR should be pretty solid. 

I've gone ahead and removed the black color from the palette, as per the notes above. Here's the full set of colors for now: 

screen shot 2018-11-19 at 1 24 06 pm

screen shot 2018-11-19 at 1 28 34 pm

The first two colors are tied to the primary color you choose in the customizer, so they'll update automatically when you update the color there:  

screen shot 2018-11-19 at 1 24 47 pm
screen shot 2018-11-19 at 1 25 16 pm

Mind giving this one more test, @allancole? I think it should be ready to go.

kjellr:
Thanks!

PR #39, opened 2018-10-16
Adds async/defer capability to JavaScript loading

mor10:
- Adds `async`/`defer` capability to JavaScript loading.
- Defers loading of `twentynineteen-skip-link-focus-fix` to improve performance.

`async`/`defer` is [new best practice for JS loading](https://developers.google.com/speed/docs/insights/BlockingJS) and should be used for all enqueued JS on the front end in place of loading in footer. This feature allows control of JS loading by adding the following code to any enqueued script:

```php
wp_script_add_data( '[label]', '[async/defer]', true );
```

Corresponds with https://core.trac.wordpress.org/ticket/12009 and gracefully falls back if/when this feature is added to core.

Code by @westonruter, originally contributed to [WP Rig](https://github.com/wprig/wprig)

PR #646, opened 2018-11-20
Add async/defer capability to JS loading. Defer menu scripts.

mor10:
- Adds async/defer capability to JavaScript loading.
- Defers loading of `twentynineteen-priority-menu` and `twentynineteen-touch-navigation` to improve performance.

[`async`/`defer`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script) is new best practice for JS loading and should be used for all enqueued JS on the front end in place of loading in footer. This feature allows control of JS loading by adding the following code to any enqueued script:

`wp_script_add_data( '[label]', '[async/defer]', true );`

Corresponds with https://core.trac.wordpress.org/ticket/12009 and gracefully falls back if/when this feature is added to core.

Code by @westonruter, originally contributed to WP Rig

Replaces #39


crunnells:
Closing this with #647.

PR #651, opened 2018-11-20
Adds default white color to link for Pullquote with solid background

### LINKED ISSUE #650: Links within solid-color pullquotes aren't visible in the editor
To reproduce: 

1. Create a solid color pullquote block
2. Add a link to that block
3. See error

pullquote-link

Note: These appear fine on the front end: 

screen shot 2018-11-20 at 9 44 51 am

mrxkon:
Adds default white color to link for Pullquote with solid background.

fixes #650 

( wp.org username: xkon )

### DIFF
--- a/style-editor.css
+++ b/style-editor.css
@@ -468,6 +468,10 @@ figcaption,
   color: #fff;
 }
 
+.wp-block-pullquote.is-style-solid-color blockquote:not(.has-text-color) a {
+  color: #fff;
+}
+
 @media only screen and (min-width: 768px) {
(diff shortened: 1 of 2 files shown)


kjellr:
Thanks, @mrxkon. In this case, we actually need the link color to be inherited rather than declared, so that it's compatible with custom text colors. Users can also technically set the background of these to be white, in which case we can't have white links. 🙂

I opened #655 with an approach that takes that into consideration, so I'll close this one for now. Thanks for the patch in any case!

PR #656, opened 2018-11-21
Try fixing social icon jumping in Safari

### LINKED ISSUE #648: Social menu icons jump on hover in Safari
hover-jumps

(Happens only on retina screens in my testing)

kjellr:
Fixes #648 

This PR applies a [hack](https://dev.opera.com/articles/css-will-change-property/#the-old-the-translatez-or-translate3d-hack) to our social media icons to prevent them from bouncing around on hover in Safari. Setting `transform: translate3d(0, 0, 0);` asks the `transition` to use hardware acceleration, which solves the jumpiness in this case. This _is_ a hack though, so I'd love to find a  better way to solve this. I tried using the new `will-change`  property here, but it didn't seem to solve the problem for me. 😕

To reproduce/test: 
- Use Safari (I used v12.0.1).
- Use a retina screen (I was only able to reproduce this on a retina screen)
- Visit a Twenty Nineteen site that has a Social Menu
- Check whether or not the icons jump on hover

**Before:**
icons-broken

**After:** 
icons-fixed

---

cc @jasmussen in case you have some insight here too.

### DIFF
--- a/sass/navigation/_menu-social-navigation.scss
+++ b/sass/navigation/_menu-social-navigation.scss
@@ -28,6 +28,9 @@
 				margin-bottom: -1px;
 				transition: opacity $link_transition ease-in-out;
 
+				// Prevent icons from jumping in Safari using hardware acceleration.
+				-webkit-transform: translate3d(0, 0, 0);
+
 				&:hover,
 				&:active {
 					color: $color__text-main;
--- a/sass/navigation/_menu-social-navigation.scss
+++ b/sass/navigation/_menu-social-navigation.scss
@@ -28,9 +28,6 @@
 				margin-bottom: -1px;
 				transition: opacity $link_transition ease-in-out;
 
-				// Prevent icons from jumping in Safari using hardware acceleration.
-				-webkit-transform: translate3d(0, 0, 0);
-
 				&:hover,
 				&:active {
 					color: $color__text-main;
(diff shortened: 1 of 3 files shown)


kjellr:
> It seems like it could be a candidate for will-change: transform;, which in the inspector also solves this issue: https://css-tricks.com/almanac/properties/w/will-change/ — this is not a hack, which is the upside.

I did take a look at this, but had trouble getting it to work. Everything I've read suggests that you should not include `will-change` on the element by default — you should only add it just before the animation will occur. So in our example, we'd want it to kick in just before the link is hovered. [The articles](https://dev.opera.com/articles/css-will-change-property/) [I've read](https://www.sitepoint.com/introduction-css-will-change-property/) all say that for hovered elements, you should try adding it when a parent is hovered. 

But... applying it to the `a` when you hover on the parent `li` doesn't work. And applying it to the `svg` when you hover on the  parent `li` doesn't work. And applying it  to the `svg` when you hover over the parent `a` doesn't work. I could only get it to work if I applied `will-change` to the `a` or `svg` at all times... which is exactly what everyone says _not_ to do. 😄 

If anyone can help get that to work, let's go for it. Otherwise, I've updated this PR to use `transform: translateZ(0);` and made that rule apply only to the `svg` — not to the entire link. This works too.

jasmussen:
In my reading, "will-change" is simply a rule that tells the computer to render this particular element on the GPU, regardless of what that element is doing or when. Which is actually a non-hacky way of applying the translateZ hack, which does the same because only the GPU can render the depth necessary for that. 

But I'm totally okay with using the translateZ hack here, because will-change is still early, and I _have_ encountered issues with it in the past.

kjellr:
That makes sense. Since translateZ works and we're all sure of it, I'll just go ahead and merge it in for now. 👍

PR #661, opened 2018-11-23
Fix PHP Warning: Parameter must be an array or an object that implements Countable

### LINKED ISSUE #659: PHP Warning: Parameter must be an array or an object that implements Countable
```
[23-Nov-2018 11:21:04 UTC] PHP Warning:  count(): Parameter must be an array or an object that implements Countable in ...\wordpress\trunk\wp-content\themes\twentynineteen\header.php on line 37
```

I'm on WordPress Nightly Build, `5.0-RC1-43944`.

joshfeck:
Fixes #659. This branch was tested on a server with PHP 7.2. View a single post with a featured image, with WP_DEBUG set to true.

### DIFF
--- a/header.php
+++ b/header.php
@@ -34,7 +34,7 @@
 					<?php twentynineteen_post_thumbnail(); ?>
 					<?php the_post(); ?>
 					<?php $discussion = ! is_page() && twentynineteen_can_show_post_thumbnail() ? twentynineteen_get_discussion_data() : null; ?>
-					<div class="<?php echo ( ! empty( $discussion ) && count( $discussion->responses ) > 0 ) ? 'entry-header has-discussion' : 'entry-header'; ?>">
+					<div class="<?php echo ( ! empty( $discussion ) && count( $discussion->authors ) > 0 ) ? 'entry-header has-discussion' : 'entry-header'; ?>">
 						<?php get_template_part( 'template-parts/header/entry', 'header' ); ?>
 					</div><!-- .entry-header -->
 					<?php rewind_posts(); ?>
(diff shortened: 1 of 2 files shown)


header.php
```diff
@@ -34,7 +34,7 @@
 					<?php twentynineteen_post_thumbnail(); ?>
 					<?php the_post(); ?>
 					<?php $discussion = ! is_page() && twentynineteen_can_show_post_thumbnail() ? twentynineteen_get_discussion_data() : null; ?>
-					<div class="<?php echo ( ! empty( $discussion ) && count( $discussion->responses ) > 0 ) ? 'entry-header has-discussion' : 'entry-header'; ?>">
+					<div class="<?php echo ( ! empty( $discussion ) && count( $discussion->authors ) > 0 ) ? 'entry-header has-discussion' : 'entry-header'; ?>">
```
glueckpress: If the intention is to count the number of responses, why change it to counting the number of authors instead? Just dumping `count()` did the trick for me locally. Since `$discussion->responses` returns a string value (which probably should be reviewed), wrapping it in `absint()` ensures we get an integer:

```html
<div class="<?php echo ( ! empty( $discussion ) && absint( $discussion->responses ) > 0 ) ? 'entry-header has-discussion' : 'entry-header'; ?>">
```

joshfeck: I made your suggested change. Since you asked, the author count seemed more reliable at the time because `$discussion->responses` is derived from `get_comment_count()` and its return value can be filtered. 

What can happen is another plugin, like muut, will filter the count to be zero (or even a placeholder to be filled later with JS). So there's a chance a post with comments will have a filtered value be zero, but that's probably an edge case. Worse case scenario is the CSS class won't be added there.

Issue #658, opened 2018-11-23
Widget titles should NOT be H2 tags


slaFFik:
Post titles in a blog view have `H2` tag. Having the same for not a content, but just widget titles steal the value from posts. I know that this is inside `aside` tag, but I'm not sure that Google (and other less significant search engines) pay too much attention to `article`, `aside` when calculating ranking of the post.
It should be at maximum `H3`, ideally even `H4`.

![image](https://user-images.githubusercontent.com/282234/48941274-88a1d080-ef23-11e8-927f-abf8827f7682.png)

I'm on WordPress Nightly Build, `5.0-RC1-43944`.

joyously:
I agree this can be a problem. I don't know why the default for widget titles is h2. But the theme can easily change it to be h3 or h4.
Is the site title always h1? Or is it a `<p>` and the page title is h1?

LittleBigThing:
> Is the site title always h1? Or is it a `<p>` and the page title is h1?

The site title is a `<p>` on other pages than the homepage.

I would stick to h2 for the widget title, but I agree that they could be smaller visually.

LittleBigThing:
> Is the site title always h1? Or is it a `<p>` and the page title is h1?

And the page title is then `<h1>` indeed.

kjellr:
> I am checking twentyfifteen, twentysixteen, and twentyseventeen. All are using h2 for widget titles. However the font size for the h2 widget title should be smaller. Like 1rem instead of 2.25em

I think we should stick with h2, for this reason. 

I'm okay with trying a smaller widget title font size (See PR https://github.com/WordPress/twentynineteen/pull/671), but I'm not 100% sure it makes sense to squeeze that in while we're in RC.

slaFFik:
It's a new theme, not a child theme for 2015-2017. Why preserve this (imo incorrect) approach by using "they've done like that" excuse? 
Semantically - this is a mistake.

kjellr:
I appreciate the continued response here, but using a `h3` or `h4` here doesn't seem like a solid solution to me. 

To expand on my short note above: The theme is making an effort to have the post/page title be the only `h1`, which leaves us with a headings structure like this by default (if a user has added no headings to post content): 

1. `h1` Page Title
2. `h2` Widget Title

If we changed the widgets to `h3` or `h4`, it's likely that there would be no `h2` tags at all on the page, leaving us with:

1. `h1` Page Title
2. `h3` Widget Title

[...]

PR #671, opened 2018-11-26
Try: Decreased font size for widget titles

kjellr:
As suggested in #658. I think this works, but it's more of an enhancement than a bug fix, so I'm not sure it fits in now that we have the release candidate.

Before:
screen shot 2018-11-26 at 10 25 42 am

After:
screen shot 2018-11-26 at 10 24 47 am

### DIFF
--- a/sass/site/secondary/_widgets.scss
+++ b/sass/site/secondary/_widgets.scss
@@ -1,6 +1,10 @@
 .widget {
 	margin: 0 0 #{$size__spacing-unit};
 
+	.widget-title {
+		font-size: $font__size-lg;
+	}
+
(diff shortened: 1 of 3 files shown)


LittleBigThing:
Longer words get quickly cut off in the widget title if they remain this large. Would that count as a bug? :-)
It is also the case for columns. Maybe both should be 'synchronized' (widget titles with h2 titles in in columns)?

kjellr:
I've migrated this PR over to trac here: https://core.trac.wordpress.org/ticket/45865

If this gets merged in, it'll happen after this GitHub repository has been closed, so I'm going to go ahead and close this PR in the meantime.

Issue #208, opened 2018-10-20
post_thumbnail size assumes large images


joyously:
When viewing a post with a small featured image, the CSS causes the browser to upscale the image and it looks jagged and bad. It also takes up way too much space (it's a vertical image).
I think that the CSS should never use width:100% on images, because the CSS can't know what size they are.

kjellr:
For visual reference, here's a screenshot using a tiny 80px by 60px featured image:

<img width="1440" alt="screen shot 2018-10-19 at 11 02 24 pm" src="https://user-images.githubusercontent.com/1202812/47250651-2705b800-d3f3-11e8-882c-f3e109c150e8.png">

nielslange:
@joyously & @kjellr On the other hand, having an 80x60 pixel featured image doesn’t look good either. Would it be a better approach to request a minimum height/width for the featured image?

joyously:
The one I had on my test site wasn't that small, but still it looked bad. And you can't enforce anything for legacy content. You could put a different class on it if it is below a threshold.
When I found this, I disabled the `width:100%` rule and the small image was repeating. It didn't look too bad that way...

kjellr:
I did some exploration around this today, and did not arrive on a great solution to this issue. Here are  some examples, each using one large image (`2520×1415`), one medium (`640x480`), and one small (`160x160`). I added some quick analysis based on my own personal impression of each option. 

## Option 1: Current

Large: 
![01-large](https://user-images.githubusercontent.com/1202812/47933954-64b31800-deac-11e8-815f-0551c89a3064.png)

Medium:
![01-medium](https://user-images.githubusercontent.com/1202812/47933956-67157200-deac-11e8-97fd-228201409841.png)

Small:
![01-small](https://user-images.githubusercontent.com/1202812/47933958-6a106280-deac-11e8-9236-71123464da01.png)

✅ Great: Large image looks ideal.
✅ Great: Medium image looks good, if not very slightly fuzzy. 
🆗 Small image is pixellated, but fills the screen.  

## Option 2: Contain

Large:
![02-large](https://user-images.githubusercontent.com/1202812/47934052-a8a61d00-deac-11e8-815a-f45fa9f29a67.png)

Medium:
![02-medium](https://user-images.githubusercontent.com/1202812/47934059-acd23a80-deac-11e8-98fa-4fde6f3f7875.png)

Small:
[...]

PR #672, opened 2018-11-26
 Try: Add a full-screen featured image toggle to the customizer

kjellr:
> ⚠️ Note: This is mostly just a proof of concept, and since there's a feature-freeze on the current Release Client, this isn't going to make it into 5.0. Even so, I wanted to share this branch as a starting point in case it becomes helpful in the future, or for anyone who may want to emulate this in a child theme or plugin. 

Currently, Twenty Nineteen displays featured images as  full-screen backgrounds on single pages:

screen shot 2018-11-26 at 12 22 06 pm

This often looks great, but can be problematic with smaller images (as noted in #208), or when images incorporate text:

screen shot 2018-11-26 at 12 23 40 pm

screen shot 2018-11-26 at 12 26 41 pm

This PR helps theme authors deal with that by offering a theme option to opt-out of the full-screen image treatment:

full-screen-image

This has only been lightly tested, but there are a few known issues:  

- The toggle lives within the Colors panel for now, so it can be tested alongside the custom colors implementation. It should likely move into a separate "Theme Options" panel.
- When the full-size image is turned off  (and color overlays are turned on), there's a slim color bar underneath the image. 
- The stylesheet updates are fairly messy. There's a lot of duplicated code for the color filters, and in general, this could use a refactor and cleanup.

### DIFF
--- a/inc/color-patterns.php
+++ b/inc/color-patterns.php
@@ -54,8 +54,10 @@ function twentynineteen_custom_colors_css() {
 		 * - WP Block Button
 		 * - Blocks
 		 */
-		.image-filters-enabled .site-header.featured-image .site-featured-image:before,
-		.image-filters-enabled .site-header.featured-image .site-featured-image:after,
+		.full-screen-image-enabled.image-filters-enabled .site-header.featured-image .site-featured-image:before,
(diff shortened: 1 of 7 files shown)


kjellr:
Closing this as per the introductory note above: This is a proof of concept and not intended to make it into the release.

PR #674, opened 2018-11-27
Add nowrap to the columns block at breakpoints above 600px

### LINKED ISSUE #673: Columns block shows non-first columns with margin-left
Non-first columns have margin-left of 30px above 600px page-width. Even though they are wrapping onto their own rows at 782px. This looks strange to my eye. I would expect all columns to be aligned once they break into 100% width and wrap. 

ezgif-5-90aca7146894

kjellr:
Fixes #673 (and also [45369 on Trac](https://core.trac.wordpress.org/ticket/45369))

Gutenberg's default front-end styles use `flex-wrap: wrap` on screens narrower than `782px`, and `flex-wrap: nowrap;` on screens above `782px`. On the frontend of Twenty Nineteen, this leads to columns wrapping earlier than we intend for them to:

columns-broken

In our case, we want to use use `flex-wrap: nowrap` on all screens above 600px wide:

columns-fixed

### DIFF
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -700,6 +700,10 @@
 			padding-right: $size__spacing-unit;
 		}
 
+		@include media(mobile) {
+			flex-wrap: nowrap;
+		}
+
 		@include media(tablet) {
 			.wp-block-column > * {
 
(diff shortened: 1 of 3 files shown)


jasmussen:
Ship it 👍 👍

PR #679, opened 2018-11-28
Remove unnecessary count() function giving warning.

### LINKED ISSUE #678: Count shouldn't be used on a numeric
Hello,

Minor issue with the use of count on a numeric, specifically the responses count used to add the .has-discussion class to the post. The use of count() throws a warning which ends up in the class in HTML;
Warning: count(): Parameter must be an array or an object that implements Countable in /app/public/wp-content/themes/twentynineteen/header.php on line 37
screen shot 2018-11-27 at 4 22 15 pm

The count function can simply be removed from here as the attribute is already an integer, the line;
https://github.com/WordPress/twentynineteen/blob/master/header.php#L40

PR coming.

Cheers

garrett-eclipse:
Remove count function as unnecessary and reduced by using concat on $classes. Fixes #678

### DIFF
--- a/header.php
+++ b/header.php
@@ -37,8 +37,8 @@
 						$discussion = ! is_page() && twentynineteen_can_show_post_thumbnail() ? twentynineteen_get_discussion_data() : null;
 
 						$classes = 'entry-header';
-						if ( ! empty( $discussion ) && count( $discussion->responses ) > 0 ) {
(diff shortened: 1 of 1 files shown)


kjellr:
Closing this, as the issue here has been solved via https://github.com/WordPress/twentynineteen/pull/661

Issue #663, opened 2018-11-24
update screenshot


joyously:
The screenshot is out of sync with the theme.
Also, it is quite large bytewise. I took the existing one and saved it as JPG, and it was 105K instead of 836K as PNG.

LittleBigThing:
And do we stick to the blog style content or will we feature a business-like page that could be built with Gutenberg?

joyously:
maybe use the starter content?

kjellr:
Thanks for pointing this out. I'm not 100% sure that starter content will make it in for 5.0 — it may be added for 5.0.1 instead. For the screenshot, I suggest we go with one of the following options: 

---

1.  An updated version of what we have currently: 

![screenshot-blog](https://user-images.githubusercontent.com/1202812/49230999-fd12be80-f3be-11e8-87d9-576adbd52a8b.png)

---

2. Or, a screenshot using the visuals that'll be in [the 5.0 About](https://core.trac.wordpress.org/ticket/45178) page (and the upcoming theme demo site I'm working on):

![screenshot-site](https://user-images.githubusercontent.com/1202812/49231118-3c410f80-f3bf-11e8-935b-042efdbfd2ec.png)

---

The first option is more visually interesting, but I lean strongly towards the second option so there's greater visual consistency in the different places people see the theme. 

> Also, it is quite large bytewise. I took the existing one and saved it as JPG, and it was 105K instead of 836K as PNG.

Both of these have been re-exported using the PNG-8 setting, [just like the screenshot in Twenty Seventeen](https://github.com/WordPress/twentyseventeen/commit/d4c496e234dfbfe38ac359b8d196f0eeaefcb03f).

LittleBigThing:
+ 1 for option 2.

I think that it's more interesting to feature things that you can do with the new editor using the theme.

Feel free to keep #216 or the PR #510 updated about ideas/plans for the starter content.

joyously:
I prefer the first one, but the readme suggests that the image might have come from pexels.com and even though it is likely licensed CC0, there has been discussion yesterday about not allowing any images from pexels. 
https://make.wordpress.org/themes/2018/11/29/pexels-images-are-no-more-allowed/

joyously:
And now someone just shared this in #themereview on Slack:
https://wptavern.com/wordpress-com-partners-with-pexels-to-offer-diverse-free-stock-photo-library

kjellr:
> there has been discussion yesterday about not allowing any images from pexels.

[...]

PR #692, opened 2018-11-29
Update theme screenshot

kjellr:
As per discussion in #663. The new screenshot is up to date with the theme and reflects the visuals from the WP 5.0 About page. It's also properly compressed, saving ~646kb.

---

**Before:** 

screenshot

---

**After:**

screenshot


allancole:
Yup, this looks good to me and looks to be closer to what users will see on activation 👍

Issue #689, opened 2018-11-29
Determine whether or not to keep two separate readme files


kjellr:
**EDIT 12/3/2018:** _We added a new `readme.txt` via #693, but we still need to figure out whether that should replace `README.md`, or whether we should keep two._

_Previous issue text:_

---

As of October 25, 2018, new themes submitted to the WordPress.org theme repository must provide a `readme.txt` in the following format: 

```
=== Theme Name ===
Contributors: (this should be a list of wordpress.org userid's)
Tags: left-sidebar, fixed-width, custom-background
Requires at least: 4.0
Tested up to: 4.3-alpha
Stable tag: 1.0.3
License: GPLv2 or later
License URI: http://www.gnu.org/licenses/gpl-2.0.html

Short description. No more than 150 chars.

== Description ==
Theme desc.

== Frequently Asked Questions ==

= A question that someone might have =

An answer to that question.

== Changelog ==

= 1.0 =
* Added new option

= 0.5 =
* Security bug addressed
* Changed thumbnail size

== Upgrade Notice ==

= 1.0 =
* Upgrade notices describe the reason a user should upgrade.  No more than 300 characters.

= 0.5 =

* This version fixes a security related bug.  Upgrade immediately.

== Resources ==
* magnify.jpg © 2014 Jane Doe, CC0
* supermenu.js © 2013-2015 James Today, MIT 
``` 

_References:_ 
https://make.wordpress.org/themes/handbook/review/required/#readme-txt-file
https://make.wordpress.org/themes/2015/04/29/a-revised-readme/

Our [current readme](https://github.com/WordPress/twentynineteen/blob/master/README.md) is in markdown format, which is standard for GitHub. We could either: 

- Reformat + rename our current readme to follow the new format (this will likely result in a hard-to-read readme for the GitHub repo, as new lines won't be preserved).
- Keep our current readme for Github, and add a new `readme.txt` file using this new format.

allancole:
#693 adds a new `.txt` reademe file and updates the formatting and description. 

Moving this issue to the 5.0.1 milestone so that we can make a decision about either keeping two separate readme files in 2 different formats, or combining them and breaking the format for either TRT requirements or GitHub legibility. 

We also need to address wether we should include an FAQ section in the readme or not.

kjellr:
This one will be taken care of in #710 — it keeps both readme files, but puts their content in sync.

kjellr:
Closing this, since #710 has been merged. 👍

PR #693, opened 2018-11-29
Add readme.txt

kjellr:
Here's a first pass at the `readme.txt` as per the requirement in #689.

A few notes: 

- [ ] I didn't touch `README.md` yet. We need to figure out whether we want to have this replace that file or not.  
- [x] For the description, I adapted text used [in the About page](https://core.trac.wordpress.org/ticket/45178). This should undergo a copy review before merge. 
- [ ] I omitted the FAQ section for now, but we should figure out if this needs one.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -18,4 +18,5 @@ Our 2019 default theme is designed to show off the power of the block editor. It
 * Initial Release
 
 == Resources ==
-* normalize.css, © 2012-2016 Nicolas Gallagher and Jonathan Neal, MIT
+* normalize.css, © 2012-2018 Nicolas Gallagher and Jonathan Neal, MIT
+* Underscores, © 2012-2018 Automattic, Inc., GNU GPL v2 or later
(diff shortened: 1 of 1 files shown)


kjellr:
@michelleweber Thanks for the feedback!

> Should it include a link to something that talks about the block editor / a demo page for people who are behind the times?

We can't right here, unfortunately — it has to be just plain text. But there will be a theme demo link, plus since the theme arrives alongside 5.0, there'll be a lot of introductory text about the block editor on [the WordPress 5.0 welcome screen](
https://core.trac.wordpress.org/ticket/45178). 

How about something like this: 

> Our 2019 default theme is designed to show off the power of the block editor. It features custom styles for all the default blocks, and is built so that what you see in the editor looks like what you'll see on your website. Twenty Nineteen is designed to be adaptable to a wide range of websites, whether you’re running a photo blog, launching a new business, or supporting a non-profit. Featuring ample whitespace and modern sans-serif headlines paired with classic serif body text, it's built to be beautiful on all screen sizes.

michelleweber:
Dig it! Good changes.

allancole:
Gonna go ahead and merge this in. We can iterate as needed :-)

Issue #696, opened 2018-11-30
Display of right-aligned images is inconsistent between the front and back end


kjellr:
## Description

Twenty Nineteen is designed to have right-aligned elements extend outside of the text column like so: 

<img width="1440" alt="screen shot 2018-11-30 at 3 24 47 pm" src="https://user-images.githubusercontent.com/1202812/49313055-1b0e1b00-f4b4-11e8-941c-c1b6fddcff18.png">

For image blocks, this occurs in the editor as intended. But on the front end, the image  is tucked all the way into the text column instead:

<img width="1413" alt="screen shot 2018-11-30 at 3 12 55 pm" src="https://user-images.githubusercontent.com/1202812/49312738-1006bb00-f4b3-11e8-92e9-7291f2d1f314.png">

## Details

This happens because our standard `alignright` block styles [target direct children](https://github.com/WordPress/twentynineteen/blob/3877766590b55756ca057c5766fd988b7a5cada8/sass/blocks/_blocks.scss#L65) of `.entry-content`. For most blocks (Cover, Pullquote, Galleries,etc.), that works just fine. They apply the `alignright` class to their topmost container like so: 

```
<div class="entry-content">
  <figure class="wp-block-pullquote alignright"> ... </figure>
</div>
```

By contrast, Left- and right-aligned images are wrapped in a container `div` that does not include the alignment class: 

```
<div class="entry-content">
   <div class="wp-block-image">
    <figure class="alignright"> ... </figure>
  </div>
</div>
```

This prevents the image from picking up our alignment styles. 

##  Possible Solutions

1. A change like https://github.com/WordPress/gutenberg/pull/10394 in Gutenberg would fix this by bringing the `alignright` classname to the wrapper `div`. This would also make this behavior more consistent with the way that classname appears for other blocks. (There's some background in there regarding why that may not be a good idea though)
2. On our side, we could adjust the editor style so that it keeps the floated item within the text column. This would at least bring consistency to what the user sees in the front and back end. This would be  a  simple change, but it'd still be a little weird that all other right-aligned blocks will float outside of the text column. 😕

---

cc  @jasmussen for some thoughts, since you know a lot of the background on this one. 🙂

Note: that this issue a little similar to #688, which occurs because non-captioned images are wrapped in a `p` tag when inserted via the Classic Editor.

kjellr:
[...]

PR #702, opened 2018-12-03
Try pushing right-aligned images outside of the text column

kjellr:
As noted in #696, right-aligned images are supposed to float outside of the text column in both the front and back end. Currently, they're not behaving as designed on the front end. 

This PR sets the `max-width` for `wp-block-image` to 100%, allowing us push right-floated images outside the content area. It required a little adjustment to get center-aligned images to work correctly, but it seems to be pretty solid in my testing. 

When testing this PR, please be sure to test all image alignments just in case. 🤞 

**Before**
screen shot 2018-12-03 at 10 14 54 am

**After**
screen shot 2018-12-03 at 10 14 05 am

### DIFF
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -509,15 +509,24 @@
 		}
 
 		.aligncenter {
+			
+			@include postContentMaxWidth();
 
 			@include media(tablet) {
-				position: relative;
-				right: calc( #{$size__site-tablet-content} / 2 );
-				transform: translateX( 50% );
+				margin: 0;
+				width: $size__site-tablet-content;
+
+				img {
+					margin: 0 auto;
+				}
 			}
 
 			@include media(desktop) {
-				right: calc( #{$size__site-desktop-content} / 2 );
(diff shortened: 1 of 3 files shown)


jasmussen:
Confirmed! This fixes the issue for me:

screenshot 2018-12-04 at 09 12 40

screenshot 2018-12-04 at 09 12 44

I also tested floats for other blocks, cover, video, wide, fullwide, a bunch of other stuff and don't see regressions. Code looks good too. 

Here's the classic editor with a vanilla right float:

screenshot 2018-12-04 at 09 17 37

screenshot 2018-12-04 at 09 18 03

Here's with a caption:

screenshot 2018-12-04 at 09 16 43

screenshot 2018-12-04 at 09 16 52

I think this is definitely shippable. Nice work!

kjellr:
Thanks for the thorough test, Joen!

kjellr:
This PR has been ported over to Trac: 
https://core.trac.wordpress.org/ticket/45716

kjellr:

PR #703, opened 2018-12-04
Ensure default outline style button has legible text on hover

rickalee:
Currently on hover .is-style-outline the button text is white on white. This checks that if no background color is set that it uses same hover color as border. 

Before & After:
outline_button-hover


kjellr:
This PR has been ported over to GitHub here: 

https://core.trac.wordpress.org/ticket/45726

laurelfulford:
This PR was committed to Core in https://core.trac.wordpress.org/changeset/44370.

kjellr:

PR #707, opened 2018-12-06
add wikipedia social icon

Klaas-:
https://core.trac.wordpress.org/ticket/45328


laurelfulford:
I'm going to close this PR, since the issue and patch have already been moved over to Core, and we're currently working on getting the rest of Twenty Nineteen's issues and patches there and closing this repo. :)

Klaas-:
@laurelfulford does that mean I have to make another patch against core or will you take this into core with other issues/patches for twenty nineteen?

kjellr:
Thanks @Klaas- I've ported this patch over to Trac here: https://core.trac.wordpress.org/ticket/45943

Klaas-:
@kjellr thanks :) btw I completely agree with you "limitations around these sorts of additions". My thought was to make this different: maybe you could attach custom logos/link to logos[privacy concerns] as needed for your site instead of using a predefined set. So instead of just creating a social link you need to add a link and an icon. That way you can accommodate everyone

Issue #717, opened 2018-12-08
Touch events hijacked by accidental JS leak


anevins12:
The 'toggleSubmenuDisplay' function in 'touch-keyboard-navigation.js' contains an event that watches and manipulates _all_ touch events and not just those specific to the submenu.

This causes a problem for links, as the code tells the browser to go to the page destination on touch - meaning people need only try to scroll the page before getting redirected to the link destination. 

https://wordpress.org/support/topic/links-get-triggered-while-scrolling-on-mobile-devices/?view=all

kjellr:
Closed via #721. Thanks!

PR #721, opened 2018-12-10
#717: Removed redundant link behaviour in touchevents js

anevins12:
AFAIK there's no need to say "If link, send the user off to the link location", because that's default behaviour of a link. Likewise there's no need to prevent that behaviour, and adding this preventdefault actually causes a console error momentarily.

### DIFF
--- a/js/touch-keyboard-navigation.js
+++ b/js/touch-keyboard-navigation.js
@@ -219,19 +219,9 @@
 
 				var url = event.target.getAttribute( 'href' ) ? event.target.getAttribute( 'href' ) : '';
 
-				// If there’s a link, go to it on touchend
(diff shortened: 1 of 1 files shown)


anevins12:
Resolves #721

kjellr:
Also: this PR takes care of [45510](https://core.trac.wordpress.org/ticket/45510) on Trac as well.

kjellr:
This fix has been ported over as a patch on Trac: 

https://core.trac.wordpress.org/ticket/45510

laurelfulford:
This PR has been committed to core in https://core.trac.wordpress.org/changeset/44358.

kjellr:

Issue #720, opened 2018-12-09
Menus: Long menus can get cut off by content when expanded


kjellr:
<img width="755" alt="screen shot 2018-12-09 at 2 49 32 pm" src="https://user-images.githubusercontent.com/1202812/49702864-48994980-fbc3-11e8-8752-6ca7e86e2c80.png">

kjellr:
We're in the process of closing down this GitHub repository now that the theme is merged into core, so I've migrated this issue over to Trac:

https://core.trac.wordpress.org/ticket/45901

I'm closing this issue a result. Please feel free to follow up in the new thread moving forward.

PR #724, opened 2018-12-10
Remove hidden overflow from site-header

allancole:
- Prevents long sub-menus from getting truncated by the header height
- Fixes: #720 

Before: 
screen shot 2018-12-09 at 2 49 32 pm

After: 
image

### DIFF
--- a/sass/site/header/_site-featured-image.scss
+++ b/sass/site/header/_site-featured-image.scss
@@ -2,9 +2,6 @@
 
 .site-header.featured-image {
 
-	/* Hide overflow for overflowing featured image */
-	overflow: hidden;
-
 	/* Need relative positioning to properly align layers. */
 	position: relative;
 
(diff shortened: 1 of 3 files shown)


LittleBigThing:
Hi,

@kjellr and @allancole: This might be trickier than it looks. This `overflow: hidden` was introduced in #450 to ensure that the background image is not visible outside of the `header`. Removing it seems to break things in Edge and IE11.

I would have to look into this in more detail to find a solution for all browsers... any ideas?

kjellr:
Thanks, @LittleBigThing — I'll remove the `approved for merge` label until we double check that.

kjellr:
@allancole — As per @LittleBigThing's notes above, this PR still had issues in IE, so I'm not going to migrate it over to Trac. The related issue is over here now: 

https://core.trac.wordpress.org/ticket/45901

Let's follow up in the new thread moving forward. Thanks!

PR #729, opened 2018-12-12
Use a button (instead of span) for the back button in the touch menu

### LINKED ISSUE #714: Navigation submenu back-button not semantic
he main nav submenu back button is implemented as a <span> tag instead of a semantic button. This means all of the useful information of a button is lost.

Solutions:
1) Use a button
2) Emulate all the button behaviour using ARIA

kjellr:
In the touch menu, the back button is currently a `span` tag. This PR changes it to a `button` for better semantics. 

Resolves #714.


kjellr:
>  I’ve got another PR on the way that sorts this out all at once. I’m still working on it but we may want to close this since it proposes this same solution.

Ok —  once that lands, we can go ahead and close this one.

kjellr:
Closing in favor of #735

PR #733, opened 2018-12-13
Remove HTML from translation strings

### LINKED ISSUE #725: [i18n] Remove unnecessary HTML from translation string
The string **`<span class="meta-nav">Published in</span><br><span class="post-title">%title</span>`** have too much unnecessary HTML that can perfectly be removed.
https://github.com/WordPress/twentynineteen/blob/master/image.php#L87

Should be something like **`Published in %s`** instead.

kjellr:
Resolves #725 (and [45473](https://core.trac.wordpress.org/ticket/45473) on Trac).

We can't actually remove the HTML from here completely because we want to maintain the visual appearance: 

screen shot 2018-12-13 at 10 05 26 am

This uses the same approach proposed by `mukesh27` on Trac: https://core.trac.wordpress.org/attachment/ticket/45473/45473.patch

I'm not sure this needs inline translator comments or not.

### DIFF
--- a/image.php
+++ b/image.php
@@ -84,8 +84,7 @@ get_header();
 					// Parent post navigation.
 					the_post_navigation(
 						array(
-							'prev_text' => '<span class="meta-nav">' . __( 'Published in', 'twentynineteen' ) . '</span> ' .
-								'<span class="post-title">%title</span>',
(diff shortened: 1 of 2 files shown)


grappler:
We can not split a string that that as in some languages the string would "Title published in". Related PR: #734

kjellr:
Closing this PR, since it needs revision and will be handled on Trac moving forward: https://core.trac.wordpress.org/ticket/45473

PR #734, opened 2018-12-13
Remove unnecessary HTML from translation string

Rahmon:
Remove unnecessary HTML from translation string: `<span class="meta-nav">Published in</span><br><span class="post-title">%title</span>`

Closes: #725

### DIFF
--- a/image.php
+++ b/image.php
@@ -84,7 +84,8 @@ get_header();
 					// Parent post navigation.
 					the_post_navigation(
 						array(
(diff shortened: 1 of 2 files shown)


grappler:
We can not split a string that that as in some languages the string would "Title published in".

Related discussion: https://github.com/WordPress/twentynineteen/pull/274

kjellr:
We're in the process of closing this repository and migrating over to the core WordPress Trac, so I'm going to close this PR in favor of the discussion happening here: 

https://core.trac.wordpress.org/ticket/45473

Please feel to weigh in on that thread! We could use your expertise.

PR #753, opened 2018-12-28
This fixes multiple js errors

jmau111:
fixes errors in priority menu and touch-keyboard-navigation js => querySelector of null


kjellr:
Hi @jmau111! Thanks for the PR. We're in the process of closing down the GitHub repository and migrating over to Trac. I've opened an issue related to some of these JS errors here: 

https://core.trac.wordpress.org/ticket/45903

I pointed readers back to this PR as a possible solution, but I'm going to close this at the moment, since we won't be doing any more merges from this repository. 

Thank you for your contribution!

jmau111:
@kjellr oh thanks for that, I missed that point, GitHub is more handy but I understand why you are doing that.

PR #754, opened 2018-12-29
Change default link SVG icon size to 26

mukeshpanchal27:
In Social Menu Walker system use SVG icon size with 26 but in default link SVG icon it will use font size 24  instead of 26 that needs to update to 26.


crunnells:
The `26` appears to be an error. The CSS overrides the height and width attributes _anyway_, and displays the icons at 32px. I think the path forward is to just _remove_ the 26, and I've created a ticket in core: https://core.trac.wordpress.org/ticket/45950

Since we're closing out this repo to move to Core Trac, I'm going to close this PR. Thanks!

PR #759, opened 2019-01-14
Syncing latest changes from Trac

allancole:
This PR simply brings in the latest changes from Trac, so that both repos are in sync before archiving this one. 

See latest changes here: https://codex.wordpress.org/Twenty_Nineteen_Theme_Changelog


kjellr:
Removed the build status from the readme too, now that Travis is turned off. This should be good to go. 👍

kjellr (approved):

